Issue #10, opened 2015-09-09
Request fails when using application/graphql with body parser


AJ1310:
When sending a Content-Type : application/graphql with a query string the request fails.
If using body parser json or url encoded it says that i don't have a query string.
if used with text type application/graphql it gets stuck on parseBody.js in the line 119.

philostler:
+1 I'm experiencing exactly the same issue as @AJ1310 describes

Aweary:
+1 I'm seeing this same issue and I've traced the hang all the way down to the `raw-body` package used to parse the `HTTPIncomingRequest` object.

`raw-body` attaches event listeners to the passed stream, but the stream is not emitting any of those events.

@leebyron [Here's is a logged copy of the stream object](https://gist.github.com/Aweary/aac594b907e12a91c285) for the query `{ post { title } }`

leebyron:
This is now fixed and will go out with the next release.

joanrodriguez:
Are you sure this has been fixed? I'm running  `express-graphql 0.4.13` with `body-parser 1.15.0` and doing a POST with headers set to `Content-Type: application/graphql` I get 

```
{
  "errors": [
    {
      "message": "Must provide query string."
    }
  ]
} 
```

joanrodriguez:
I solved my problem doing the following:

```
// configure app to use bodyParser()
// this will let us get the data from a POST
// exclusing the route to graphql
app.use(/\/((?!graphql).)*/, bodyParser.urlencoded({ extended: true }));
app.use(/\/((?!graphql).)*/, bodyParser.json());
```

leebyron:
Yes this is fixed. However if you're also using body-parser, you must still ensure the output of body-parser is compatible with the express-graphql plugin. 

The recommended use is still to use the express-graphql middleware *before* using any kind of body parser middleware you may be using for other routes.

AvraamMavridis:
I am experiencing the same issue using :

    "body-parser": "^1.17.2",
    "express": "^4.15.3",
    "express-graphql": "^0.6.


My server:

```
const schema = require('./schema');
const express = require('express');
const graphqlHTTP = require('express-graphql');
const cors = require('cors')
const bodyParser = require('body-parser');


const app = express();
app.use(cors());
app.use(bodyParser.urlencoded({ extended: true }));
app.use(bodyParser.json());

app.use('/graphql', graphqlHTTP(req => {
  return ({
  schema,
  pretty: true,
  })
}));

[...]

PR #11, opened 2015-09-17
Check for req.body parsed as a string

Aweary:
`req.body` may be a string if parsed with something like `body-parser`. If it is, this creates the proper object shape used by `getGraphQLParams`

Fixes https://github.com/graphql/express-graphql/issues/10

### DIFF
--- a/package.json
+++ b/package.json
@@ -58,6 +58,7 @@
     "babel-core": "5.8.22",
     "babel-eslint": "4.0.10",
     "babel-runtime": "5.8.20",
+    "body-parser": "^1.14.0",
     "chai": "3.2.0",
     "coveralls": "2.11.4",
     "eslint": "1.1.0",
--- a/src/__tests__/http-test.js
+++ b/src/__tests__/http-test.js
@@ -16,6 +16,7 @@ import { describe, it } from 'mocha';
 import { stringify } from 'querystring';
 import zlib from 'zlib';
 import multer from 'multer';
+import bodyParser from 'body-parser';
 import request from 'supertest-as-promised';
 import express4 from 'express'; // modern
 import express3 from 'express3'; // old but commonly still used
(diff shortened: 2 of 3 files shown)


leebyron:
Awesome, thank you for digging into this! Would you mind also adding a test similar to and next to https://github.com/graphql/express-graphql/blob/master/src/__tests__/http-test.js#L461 which illustrates the repro for #10? That will ensure this never regresses!

leebyron:
Another question, in the case that a body has already been parsed have request headers been checked as well? The existing parsed-object query expects request properties in a form that express-graphql expects, but when parsed as a string how can we ensure the request was intended to be a posted query? Perhaps in this condition we should check the content type header to see if `application/graphql` was sent, and only fulfill as a query in that case?

Aweary:
The request header wouldn't have been validated at that point. If we only want to respond to pared requests with a content-type of `application/graphql` then we do something like

```js
if (typeof req.body === 'string') {
      return req.headers['content-type'] === 'application/graphql'
            ? next(null, { query: req.body })
            : next(new Error('...'));
  }
```

Assuming you want to pass an error on to `next` if the header isn't correct.

leebyron:
I think we should mirror the current behavior of checking for the header and otherwise parsing nothing (no error). Ideally we can recycle `typeInfo` by moving the check you're doing to below that line.

Aweary:
Got it, I moved the check below `typeInfo` and used that to determine the return value. I also added that unit test. I tried to keep it as close to the other tests as possible, but let me know if its not up to snuff.

Aweary:
@leebyron any word on this?

leebyron:
I just squashed your refs and merged in e6fb666f87bddd383dde571f89871b2dd66f3006

Thanks for spotting this!

PR #16, opened 2015-09-29
[RFC] GraphiQL when loaded in browser

leebyron:
This RFC adds an opt-in behavior of express-graphql to render a fully functional GraphiQL instance when loaded from a browser - determed as anything that prefers HTML over JSON from the Accept header.

GraphiQL mode is enabled by providing `graphiql: true` in the options bag. It defaults to `false` which is a reasonable default, but also ensures any existing use updating to this version will not unknowingly add this feature. The examples in the README have been changed to enable GraphiQL, ideally all future users will leave this on, cargo-cult from README is likely.

When GraphiQL is shown, the query result is run on the server first and presented immediately in the GraphiQL UI.

If you wish to use the browser as a cheap cURL, just add `&raw` to the end of the URL and GraphiQL is replaced by raw JSON output.

One small detail: any non-GraphQL GET params provided will be carried through by GraphiQL when executing queries. Ideally this covers the majority of authentication cases which are likely to look something like `https://mysite/graphql?oauth=abc123&query={yo}`.

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -12,6 +12,7 @@ import httpError from 'http-errors';
 import { graphql } from 'graphql';
 import { formatError } from 'graphql/error';
 import { parseBody } from './parseBody';
+import { renderGraphiQL } from './renderGraphiQL';
 import type { Request, Response } from 'express';
 
 /**
@@ -34,6 +35,11 @@ export type OptionsObj = {
    * A boolean to configure whether the output should be pretty-printed.
    */
   pretty?: ?boolean,
+
+  /**
+   * A boolean to optionally enable GraphiQL mode
+   */
+  graphiql?: ?boolean,
 };
 
 type Middleware = (request: Request, response: Response) => void;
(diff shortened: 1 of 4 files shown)


leebyron:
This Good Idea™ comes from @steveluscher

src/renderGraphiQL.js
```diff
@@ -0,0 +1,123 @@
...
+  var resultString =
+    data && data.result ? JSON.stringify(data.result, null, 2) : null;
+
+  /* eslint-disable max-len */
+  return `<!--
+The request to this GraphQL server provided the header "Accept: text/html"
+and as a result has been presented GraphiQL - an in-browser IDE for
+exploring GraphQL.
+
+If you wish to receive JSON, provide the header "Accept: application/json".
```
dschafer: Mention `raw` here.

src/index.js
```diff
@@ -152,6 +175,17 @@ function getGraphQLParams(request: Request, data: Object): GraphQLParams {
 }
 
 /**
+ * Helper function to determine if GraphiQL can be displayed.
+ */
+function canDisplayGraphiQL(request: Request, data: Object): boolean {
+  // If `raw` exists, GraphiQL mode is not enabled.
+  var raw = request.query.raw !== undefined || data.raw !== undefined;
+  // Allowed to show GraphiQL if not requested as raw and this request
+  // prefers JSON over HTML.
```
dschafer: This is flipped

dschafer:
Looks good. Ship it.

PR #18, opened 2015-09-30
Disable GET for non-query operations

### LINKED ISSUE #17: Opt-in to accepting mutations on GET
Performing mutations on an http GET request is typically bad behavior. It's usually only presumed safe to do this if providing some single-use token to avoid CSRF attacks.

express-graphql should not execute mutations on GET requests unless some opt-in is provided.

leebyron:
This fixes #17 by disabling GET requests for non-query operations. However GraphiQL may still be presented for these GET queries, but the query will just not be immediately issued, allowing for exploration of mutations.

I considered adding an option flag to enable mutations on GET, but decided that this edge case is sufficiently rare that we should only introduce it when there is a clear and compelling use case so that we get the API correct.

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -9,8 +9,11 @@
  */
 
 import httpError from 'http-errors';
-import { graphql } from 'graphql';
 import { formatError } from 'graphql/error';
+import { execute } from 'graphql/execution';
+import { parse, Source } from 'graphql/language';
+import { validate } from 'graphql/validation';
+import { getOperationAST } from 'graphql/utilities/getOperationAST';
 import { parseBody } from './parseBody';
 import { renderGraphiQL } from './renderGraphiQL';
 import type { Request, Response } from 'express';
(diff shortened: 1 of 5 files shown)


src/index.js
```diff
@@ -186,6 +231,31 @@ function canDisplayGraphiQL(request: Request, data: Object): boolean {
 }
 
 /**
+ * Helper function to produce an operation AST given a document and an
+ * operation name.
+ */
+function getOperation(documentAST: Object, operationName: ?string): ?Object {
```
dschafer: Should this live in `graphql-js` somewhere?

dschafer:
This looks good, ship it.

src/index.js
leebyron: I had the same thought. I'll do that but give it some time to replace this usage to avoid bleeding-edge dependency-hell

leebyron: https://github.com/graphql/graphql-js/commit/f66ddf55f93010f7e2ca4ca717d522d4d0d4b12b

PR #25, opened 2015-10-12
Set `credentials` to `include` in graphQLFetcher

### LINKED ISSUE #24: GraphiQL Auth
Set fetcher to use `withCredentials` to allow cookies to be sent for auth.

cgarvis:
Should fix #24


cgarvis:
@leebyron thoughts?

vslinko:
I really want to this PR be merged cause my mutations depends on user in session.

leebyron:
Sorry for the delay! This is great. I'll have a release out later today. This is more or less a copy from the graphiql repo. Would you mind applying the same change to the examples file in that project as well?

PR #27, opened 2015-10-16
Fix linter errors and warnings from resources files

brenoc:
The title says it all :)


leebyron:
I'm not sure how you encountered this, because we don't run the linter on this directory, but sure why not :)

PR #28, opened 2015-10-23
allow session cookies on same origin

cbosuna:
I was trying to pass session information as a root value.  Without this change it doesn't work through graphiql.


leebyron:
Thanks! This same change landed in master a few weeks ago and has been live with the latest version on npm.

PR #31, opened 2015-10-27
Update GraphiQL cdn version

mjw56:
Was fiddling around in chrome dev tools while using the GUI and noticed that the cdn version was not the latest. So this PR will bring it up to date.


leebyron:
Thanks! Now updated to the latest

Issue #36, opened 2015-11-20
A way to post-process the response


mnylen:
Would be nice if there was some way of post-processing the response. Some use cases:

* Log errors
* Sanitize error messages from any sensitive information
* Return 400 on validation error; 500 on any other error
* Add custom response headers based on response body

mnylen:
A nasty hack that I'm currently using to achieve this:

```
app.use('/graphql', function(req, res, next) {
  let end = res.end

  res.end = function(chunk, encoding) {
    res.end = end

    let contentType = res.get('content-type') || 'text/html'
    if (contentType.indexOf('application/json') === 0) { // don't overwrite graphiql responses
      let response = JSON.parse(chunk)

      if (response.errors) {
        let validationErrors = response.errors
          .filter(({message}) => message.indexOf('Syntax Error') !== -1 || message.indexOf('Cannot query field') !== -1)
          .map(({message}) => ({ message }))

        if (validationErrors.length > 0) {
          res.status(400).json({ errors: validationErrors })
        } else {
          res.status(500).json({ errors: [{ message: "Query could not be executed" }] })
        }
      } else {
        res.end(chunk, encoding)
      }
    } else {
      res.end(chunk, encoding)
    }
  }

  return next()
})
```

mnylen:
Still, although this hack allows me to sanitize the error messages, it doesn't give me an easy way to access the parsed query and variables from request body / query string.

Ideally I'd like to do something like:

```
app.use('/graphql', graphqlHTTP({
  schema: schema,
  graphiql: true,
  writeResponse: (query, variables, result, req, res) => {
    if (result.errors) {
      logger.error("Error executing query: " + query + "\nVariables:\n" + JSON.stringify(variables) + "\nErrors:\n" + JSON.stringify(result))
      res.send(500).json(sanitizeErrors(result))
    } else {
      res.send(200).json(result)
    }
  }
})```

leebyron:
Those use cases sound like things we should support, but post-processing the response is not how I would prefer them implemented.

The goal of this package is to create a standardized http interface for GraphQL, which post-processing could easily circumvent.

I'll propose some paths forward:

> Log errors

graphql-js could already use a comprehensive logging interface, regardless of if it is accessed via http. I'd like to focus a logging interface on graphql-js itself, and just make that easy to interface via express-graphql.

> Sanitize error messages from any sensitive information

Similar to the above. Sanitizing sensitive info in errors seems like an issue regardless of the access mechanism. 

[...]

PR #37, opened 2015-12-02
Add support for errorsCallback so we got errors detail on server …

ohardy:
I added a test for this but sinon must be used for this test to really work. Didn't know if you are OK to include sinon and sinon-chai ?

### DIFF
--- a/package.json
+++ b/package.json
@@ -66,7 +66,7 @@
     "express": "4.13.3",
     "express3": "*",
     "flow-bin": "0.18.1",
-    "graphql": "0.4.8",
+    "graphql": "0.4.14",
     "isparta": "3.0.3",
     "mocha": "2.2.5",
     "multer": "1.0.3",
--- a/package.json
+++ b/package.json
@@ -40,6 +40,7 @@
     "lint": "eslint src",
     "check": "flow check",
     "build": "rm -rf dist/* && babel src --ignore __tests__ --out-dir dist",
+    "postinstall": "rm -rf dist/* && babel src --ignore __tests__ --out-dir dist",
     "watch": "babel --optional runtime resources/watch.js | node",
     "cover": "babel-node node_modules/.bin/isparta cover --root src --report html node_modules/.bin/_mocha -- $npm_package_options_mocha",
     "cover:lcov": "babel-node node_modules/.bin/isparta cover --root src --report lcovonly node_modules/.bin/_mocha -- $npm_package_options_mocha",
--- a/package.json
+++ b/package.json
@@ -21,7 +21,11 @@
     "api"
   ],
   "main": "dist/index.js",
+  "directories": {
+    "lib": "./dist"
+  },
   "files": [
+    "src",
     "dist",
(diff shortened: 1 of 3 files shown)


leebyron:
Two primary points of feedback:

First, please ensure this change is well scoped. This makes some seemingly unrelated changes to the package.json file.

Second, could you help me understand the motivation for this change? This seems directionally good, but I'd like to better understand what you're trying to accomplish by proposing this, since there may be other solutions to consider.

leebyron:
Also please see and mirror existing test formatting rather than including new test mechanisms

cdonohue:
This could be extremely helpful.

If `graphql` is hooked up to a database to make a query, wouldn't you want to intercept any errors and handle them accordingly? Currently, you are only able to format the errors, which passes them in a `200` response with the errors inside an `errors` property.

And, when promises are used, every request will go through as a success _unless_ there is a `graphql` [syntax error](https://github.com/graphql/express-graphql/blob/master/src/index.js#L142), [wrong verb used](https://github.com/graphql/express-graphql/blob/master/src/index.js#L160), etc.

So, in database example above, How would one go about sending back anything other than a `200`?

cdonohue:
@leebyron @ohardy ?

ohardy:
@cdonohue For me it's OK now with `formatError`, not you ?

cdonohue:
I guess I'm just looking for a solution where errors can be caught _and_ the response be set accordingly.

Take the example above. If the graphql backend is connected to a database and for some reason, the database is down (problem outside of graphql). When the graphql query/mutation returns, it will still send a `200` status back to the client. You can format the errors all you want, but you are still dealing with a `SUCCESS`.

This also means that you would have to handle your client requests a bit differently for `express-graphql`:
```js
axios.post("/path-to-graphql", { query, variables })
  .then(response => {
    if (response.data.errors && response.data.errors.length) {
      throw response.data.errors[0];
    }
    // --> process data
  })
  .catch(error => {
    // handle client errors
  })
;
```

To me, the `catch` should do it's job _without_ having to check `response.data.errors' and throw accordingly.

Maybe this should go in its own issue? But I already saw that https://github.com/graphql/express-graphql/issues/36 has been closed out.

How are others handling this case when errors happen on the server and `200` status codes are being sent?

leebyron:
Returning 200 is intentional in the case that a nullable portion of a query failed, but other parts of the request succeeded. Express-graphql should return an error code if the whole query fails

cdonohue:
Thanks for clarifying, @leebyron. I'll revisit my example and double-check everything.

leebyron:
Closing this as I think we have better solutions for the two motivating examples:

1. You wish to format error output before they get to the client. Options can include a `formatError` function which does exactly this.
2. You wish to log errors as they occur for metrics purposes. express-graphql is probably the wrong layer of abstraction to do such a thing, this isn't implemented yet, but there is a tracking task for this very purpose at https://github.com/graphql/graphql-js/issues/284.

PR #43, opened 2016-01-29
show user-friendly errors in graphiql if request fails to server for any reason

chmontgomery:
I was seeing very unfriendly errors in graphiql if there were issues connecting to the underlying server for any reason, e.g.  we front all our apps with nginx and if the request took too long the user would see this:

screen shot 2016-01-29 at 2 13 43 pm

First, I updated the `fetch` chain in `renderGraphiQL.js` to [check status codes per the fetch.js best practices](https://github.com/github/fetch#handling-http-error-statuses) before trying to parse the response json. Second, I catch all errors and return a graphql friendly error message. Now my users will see this:

screen shot 2016-01-29 at 2 43 28 pm

### DIFF
--- a/src/renderGraphiQL.js
+++ b/src/renderGraphiQL.js
@@ -90,8 +90,24 @@ add "&raw" to the end of the URL within a browser.
         },
         body: JSON.stringify(graphQLParams),
         credentials: 'include',
+      }).then(function checkStatus(response) {
+        if (response.status >= 200 && response.status < 300) {
+          return response;
+        } else {
+          var error = new Error(response.status + ' (' + response.statusText + ')');
+          error.response = response;
+          throw error;
+        }
       }).then(function (response) {
         return response.json();
+      }).catch(function (err) {
+        return {
+          "errors": [
+            {
+              "message": err.message
+            }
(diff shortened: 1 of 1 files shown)


leebyron:
This is cool, nice work!

A few thoughts. First, there are plenty of times when we return a non-200 response, but we are also sending well formed JSON bodies. Examples include syntax or validation errors. An example in the code: https://github.com/graphql/express-graphql/blob/master/src/index.js#L170

We want to ensure that these are not malformed.

I think the condition you want to look for is not that the response is non-200, but that the response failed to return valid JSON. In that case you likely want to just dump the response directly into the pane instead of trying to JSON encode it, otherwise you may accidentally misinterpret your client-side JSON form as what the server sent you, obfuscating the server's issue.

leebyron:
@chmontgomery please give the latest version of `express-graphql` a test. I think b9f872198ca514bfa876df4715277a83c7568e57 - a change synced from the recent `graphiql` updates will likely solve your issue. Please let me know if it does not.

leebyron:
Closing this as it was mitigated in a prior release.

PR #48, opened 2016-02-20
Support passing a promise as the schema option

sorenbs:
I need to load the schema from a database based on request headers. As far as I can figure there is no way to do this currently. In this pr I support passing a promise to the schema option. Is this a good approach?

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -87,22 +87,23 @@ export default function graphqlHTTP(options: Options): Middleware {
     new Promise((resolve, reject) => {
 
       // Get GraphQL options given this request.
-      const optionsObj = getOptions(options, request);
-      schema = optionsObj.schema;
-      rootValue = optionsObj.rootValue;
-      pretty = optionsObj.pretty;
-      graphiql = optionsObj.graphiql;
-      formatErrorFn = optionsObj.formatError;
-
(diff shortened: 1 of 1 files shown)


leebyron:
Schema is retrieved from Options which is already a function of the request. Is there any reason you can't use that existing mechanism?

```js
app.use(urlString(), graphqlHTTP(req => {
  return {
    schema: req.header.fooBar ? SchemaOne : SchemaTwo,
  };
}));
```

sorenbs:
Thanks Lee

I need to fetch the schema from a database. My initial thought was to wrap options in a promise like this:

```
app.use(urlString(), graphqlHTTP(req => {
  return getSchema(req).then(schema => {
    return {
      schema: schema,
    };
  });
}));
```

But graphqlHTTP expects options to be an object, not a promise.

My workaround is to have an extra middleware that retrieves the schema and sticks it on the request like this:

```
app.use(urlString(), (req, res, next) => {
  getSchema(req)
  .then(schema => {
    req.schema = schema;
    next();
  })
});

app.use(urlString(), graphqlHTTP(req => {
    return {
      schema: req.schema,
    };
}));
```

This works but is a little clunky.

leebyron:
That makes sense! I've added this capability now

sorenbs:
Thanks!

PR #51, opened 2016-03-08
Add Promise fallback for node.js v0.10

evanshortiss:
I understand that Node.js v0.10.* is on the way out, but this fallback would be nice since it will only use bluebird if the current runtime does not support Promises.

Personally I work in a few environments where v0.10 is the only option at present so this would be great.

### DIFF
--- a/package.json
+++ b/package.json
@@ -46,6 +46,7 @@
     "preversion": "npm test"
   },
   "dependencies": {
+    "bluebird": "^3.3.4",
     "content-type": "~1.0.1",
     "http-errors": "~1.3.1",
     "raw-body": "~2.1.2"
--- a/src/index.js
+++ b/src/index.js
@@ -17,6 +17,7 @@ import {
   getOperationAST
 } from 'graphql';
 import httpError from 'http-errors';
+import Promise from 'bluebird';
 
 import { parseBody } from './parseBody';
 import { renderGraphiQL } from './renderGraphiQL';


leebyron:
I'd rather not do this as this would force a bluebird (non-standard Promises) dependency on all users of `express-graphql`, not just node v0.10.

Since this library expects Promise to be a global, you can pretty easily shim it and continue to use node v0.10 alongside this library or any others that require Promise. That also allows you to use whichever Promise shim you like:

```js
global.Promise = require('bluebird');

// Import express, express-graphql etc.
```

evanshortiss:
@leebyron surprisingly ```global.Promise = require('bluebird');``` didn't work, as that was my first attempted solution. Will try again as a sanity check however.

Also, Bluebird will only be used in node 0.10. In all other versions bluebird will simply return the "real" Promise implementation.

Thoughts?

leebyron:
Are you bundling files before running them or something like that? I've seen that impact the use of global before.

evanshortiss:
I stand corrected @leebyron, have just run this again and it works flawlessly as it should. I believe this was the cause:

```
var graphqlHTTP = require('express-graphql');
var graphql = require('graphql');
global.Promise = require('bluebird');
```

vs

```
global.Promise = require('bluebird');

var graphqlHTTP = require('express-graphql');
var graphql = require('graphql');
```

It was late night when I first worked on this; at least I'm sticking to that excuse!

leebyron:
Ah yes. It's definitely important to do any edits to the global before anything else. I'm glad you got things working.

PR #53, opened 2016-03-10
Custom validation rules

tuananhtd:
The most useful use case I could think of right now which also is the reason I proposed this is to set limit on query complexity. This would be a great starting point as we can take the advantage of the awesome [graphql's visitor api](https://github.com/graphql/graphql-js/blob/d5ecfe08b1f74dedd3e42186426100eddc07da7a/src/language/visitor.jsl).


leebyron:
Nice work. This is a good idea

Issue #47, opened 2016-02-11
Bad escape of query result on initial load of page


KenanSulayman:
We're running into an issue with GraphiQL that our QA team uncovered where the page won't properly render when the result of the query specified in the query-string "query" parameter includes HTML as a string and it is injected into the page on initial load:

![untitled](https://cloud.githubusercontent.com/assets/280212/12977158/6f6ce49a-d0c9-11e5-8747-0f19205ab1f1.png)

The current workaround is to copy and paste the query into a new page w/o query string or remove the query string and reload, so that the query is loaded from the localStorage.

leebyron:
Thanks for the report, but can you offer more details? For example is there a simple query that always causes this issue to reproduce? Perhaps one you can illustrate with http://graphql-swapi.parseapp.com/?

KenanSulayman:
@leebyron the results have to be values that include nested html, I'm afraid that's not the case in the StarWars-example

jamesbehr:
This is actually a reflected XSS vulnerability. You can reproduce this on the SWAPI example by using [this link](http://graphql-swapi.parseapp.com/?query=%3C%2Fscript%3E%3Cscript%3Ealert(%27XSS%27);%3C/script%3E).

Or you can input the "query" 
```html
</script><script>alert('XSS');</script>
```
and then refresh the page.

PR #56, opened 2016-03-18
Escape query, variables and response on initial page load 

jamesbehr:
HTML was was not properly escaped in queries or variables causing GraphiQL to break. See #47


leebyron:
Thanks!

PR #57, opened 2016-04-05
Add context argument to express-graphql to make it work with 0.5.0-beta.1 release of graphql-js

helfer:
This adds the context as an argument, making sure that express-graphql works with graphql-js v0.5.0-beta.1

### DIFF
--- a/package.json
+++ b/package.json
@@ -48,10 +48,8 @@
   "dependencies": {
     "content-type": "~1.0.1",
     "http-errors": "~1.3.1",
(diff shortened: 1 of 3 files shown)


leebyron:
Thanks for getting started on this! I hope to have the non-beta version of graphql-js out very soon, at which point we will definitely want to include this change here.

leebyron:
Merged in 0b53f8ae4c6297f204d2926a653023e6497eb706

PR #58, opened 2016-04-06
Move to Container Based Travis

chentsulin:


leebyron:
This project was started after travis-ci started issuing container based as a default, so this project is already container based without any additional changes to the dotfile.

PR #62, opened 2016-04-09
Use ReactDOM.render

robrichard:


istarkov:
Without this PR, GraphiQl fails on render

ericclemmons:
@leebyron Good evening!

According to commit history, you seem to be the one doing tagged releases.

Can you push out `v5.0.1` with this patch?

Thanks!

leebyron:
Thanks, sorry for the churn!

PR #64, opened 2016-04-10
Fix react rendering

calebmer:
GraphiQL currently doesn't render because `React.render` is undefined. It should probably be `ReactDOM.render`.


chentsulin:
duplicate of #62

calebmer:
Hah, didn't even check 😊 

Closing as #62 is older. Here's hoping it gets merged soon!

PR #66, opened 2016-04-10
use Promise.resolve instead of new Promise(resolve

chentsulin:


leebyron:
This would break the case where `options` function throwing an error should be caught.

I'll add a test that would have failed with this PR.

chentsulin:
Thanks for the explanation!

PR #68, opened 2016-04-10
Use ReactDOM.render

rippedspine:
graphql:104 Uncaught TypeError: React.render is not a function


chentsulin:
duplicate of #62, #64

PR #70, opened 2016-04-12
Update to React 15.0.1 and use ReactDOM to render

jepezi:
Fix React.render is not a function error by using ReactDOM.render. Also update React to 15.0.1.


jepezi:
Sorry for duplication.

PR #73, opened 2016-04-14
Add some meta to GraphiQL head

calebmer:
This adds some basic metadata to improve UX including:

- A charset meta tag. This helps in case GraphQL data contains some special UTF8 characters.
- A title like: `http://localhost:3000/?query=%7B%0A%20%20personList%20%7B%0A%20%20%20%20edges…` is not a fun user experience.
- A meta tag which prevents robots from indexing a GraphiQL page. GraphiQL should probably never be exposed publicly, however accidents do happen.


calebmer:
Sorry, didn't notice the lint error. Fixed 👍

Issue #75, opened 2016-04-19
Future plans for this repository


calebmer:
What is the roadmap/future plans for this repository? I'm working on a GraphQL HTTP server and I'm evaluating whether I should use this project or build my own. I am a little confused on whether this repository is supposed to be a reference implementation for GraphQL HTTP server authors or is this supposed to be *the* resource for building production grade HTTP GraphQL servers with GraphQL JS. If it is not the latter I ask if that might be reconsidered.

There are a few features which I would like to see before using this in production (in order of importance):

- Remove the `express` dependency. This module could easily be just a connect-style middleware which could then work with `express`, `connect`, `router`, and `koa` (off the top of my head). This might warrant a module name change to `graphql-http` or something.
- Query cache for production. These queries would be pre-parsed/pre-validated and they would be referenced using their operation name by a client.
- By default add an extra validation rule to prevent malicious queries (could be implemented in another module and imported).

These are things I would be willing to contribute (as I would be building them on my own anyway…). If I get approval from the current core team I can start developing these things. I would rather contribute to this repository over creating a new one or keeping the code close sourced.

leebyron:
> is supposed to be a reference implementation for GraphQL HTTP server authors or is this supposed to be the resource for building production grade HTTP GraphQL servers with GraphQL JS.

Definitely yes to the first. This is a reference implementation of how a GraphQL HTTP server should operate, that is the primary goal.

Usually yes to the second, but deferring to the primary goal of being a reference implementation should a change trade-off between the two.

If you're building a node service with graphql-js, then this service should enable your production traffic in the large majority of cases.

leebyron:
> Remove the express dependency. This module could easily be just a connect-style middleware which could then work with express, connect, router, and koa (off the top of my head). This might warrant a module name change to graphql-http or something.

I like this. I think this was just short-sighted on my own part, since I didn't realize these server modules all shared the same middleware API. This should probably be called `graphql-http-middleware` or something like that. Removing the peer-dep for express seems totally acceptable.

> Query cache for production. These queries would be pre-parsed/pre-validated and they would be referenced using their operation name by a client.

@alangenfeld has been thinking about how to add that to this repo in a way that is generic to how you actually want to cache those things. Once we have that we should include examples of how to use it.

> By default add an extra validation rule to prevent malicious queries (could be implemented in another module and imported).

[...]

PR #77, opened 2016-04-23
Implementation of features from #75

calebmer:
This PR contains a couple of new features (because it is easier to open one PR instead of a bunch of PRs with dependencies on one another). This contains an implementation of features from #75 and the small GraphiQL changes I made in #73.

A full list of changes is as follows:

- Updating of the eslint config (using the config from [graphql/graphql-js](https://github/graphql/graphql-js) as I assume they are one in the same).
- Support for any node server framework which supports connect style middleware.
- Feature for production servers which disallows clients sending `query` parameters instead using `operationName` and a predefined GraphQL document.
- Documentation for these features.

With these changes `express-graphql` will no longer only support `express`, so it would be nice to see the project renamed to `graphql-http` representing its polymorphism over HTTP server frameworks. If desired I can also create `./hapi.js` and `./koa.js` files for interoperability with these frameworks.

GraphQL is awesome, let's give it awesome server tooling 👍


calebmer:
Do you know why tests are failing on `iojs` and `0.12`? I can't reproduce these errors locally…

calebmer:
\cc @leebyron

leebyron:
Could you figure out how to break this into independent PRs? This is great work, but keeping changes small and scoped will be easier to review and easier to find any intermediate breakages

calebmer:
Ok, it's a little more work for me. Expect a couple of PRs after some cherry picking.

calebmer:
@leebyron split up all the PRs into #81, #82, #83, #84, and #85. I'll be fixing tests I missed as they pop up on CI in case I missed it locally 😖

When the issue I couldn't debug is isolated I'll mention it.

PR #79, opened 2016-04-30
Make dependencies up-to-date, needed to fix bug: https://phabricator.babeljs.io/T2212

bkinsey808:
So, I ran into an issue with the dist/index.js file. I was trying to consume this project as a dependency in a typescript project.  Apparently the older version of babel was incorrectly inserting

```
module.exports = exports["default"]
```

into the compiled code, which was causing this issue: https://github.com/Microsoft/TypeScript/issues/5565

PR #81, opened 2016-05-01
Update eslint config

calebmer:
I was having some troubles with developer tools because eslint was outdated, so this was a prerequisite for development. Maybe duplicate of #79/#80. This PR really isn't required for what I want to accomplish, but helpful for people looking to hack on the codebase.


leebyron:
Thanks - was waiting on this to upgrade to use babel6 to compile, but I'm fine with this as an incremental step forward.

PR #82, opened 2016-05-01
Connect support

calebmer:
General node `http` compatibility tested by `connect` support. I could also add a `./koa.js` and `./hapi.js` file at the root of this project to support those frameworks.


calebmer:
So coveralls is failing because I'm not testing `index.js` after moving its contents to `graphqlHTTP.js`. Would you prefer the contents of `graphqlHTTP.js` in `index.js` and if not, is there a way to get coveralls to pass with the change?

helfer:
@calebmer It's really hard to tell in the PR summary what you changed because all the content was moved to graphqlHTTP.js
 
I don't think the renaming change is related to connect/koa/hapi support, so if at all, it should be in a different PR.

I'm actually interested in merging this into apolloServer soon, and add koa and hapi support, so this PR will be a helpful reference for me.

calebmer:
Ok, I'll move it back. For the meantime you can see [this](https://github.com/calebmer/express-graphql/commit/d4384fac2037bd338502e16668f616f98f09173a) commit.

Out of curiosity is Apollo server forking this project or using it as a dependency?

helfer:
> Out of curiosity is Apollo server forking this project or using it as a dependency?
We started out using it as a dependency, then used a fork, but I'm finding more and more things that I think we want to do differently, so I'm planning to write a middleware from scratch, but use the test suite from express-graphql, which is really great.

src/index.js
```diff
@@ -277,8 +284,23 @@ function getGraphQLParams(request: Request, data: Object): GraphQLParams {
...
   // If `raw` exists, GraphiQL mode is not enabled.
-  const raw = request.query.raw !== undefined || data.raw !== undefined;
+  const raw = getQuery(request).raw !== undefined || data.raw !== undefined;
   // Allowed to show GraphiQL if not requested as raw and this request
   // prefers HTML over JSON.
-  return !raw && request.accepts([ 'json', 'html' ]) === 'html';
+  return !raw && accepts(request).types([ 'json', 'html' ]) === 'html';
+}
+
+function getQuery(request: Request): Object {
```
leebyron: This portion seems unrelated

leebyron:
This is awesome, thanks @calebmer, sorry for the delay getting to this.

leebyron:
Landed in 23e1a0c5abb3afbde5d6ff895cd88f81a99e0553

jaulz:
@calebmer Great work. Any hint how to integrate this with hapi now?

calebmer:
Unfortunately, for whatever reason Hapi has decided to not support connect-styled middleware [see this issue](https://github.com/hapijs/hapi/issues/80). Instead look at projects like [`hapi-graphql`](https://github.com/SimonDegraeve/hapi-graphql).

If enough people are interested though, it wouldn’t be too hard to extract the HTTP interface and allow for different implementations.

helfer:
@calebmer @jaulz I thought you might find apollo-server interesting in that case. A couple of us started this project to build a production-ready and easy to use javascript GraphQL server. Right now we're factoring out the core functionality, and are working on integrations with express, connect, HAPI and Koa: https://github.com/apollostack/apollo-server/tree/core-refactor

Some of the features we're going to add are things like support for batching, stored queries, a websocket transport etc.

PS: hapi-graphql is interesting, but we decided against using it because it doesn't do GraphQL validation at all (a security issue) and the author doesn't seem to be maintaining it at all.

calebmer:
@helfer what’s your plan for stored queries? My implementation is here: #83.

helfer:
The idea is to have a [core](https://github.com/apollostack/apollo-server/tree/core-refactor/src/core) which accepts either a query string or a document. If it's a document, it skips validation. Then there should be a query store, in which queries can be put at startup after running validation. After that it's up to the integration (express, HAPI, connect, Koa) to get the query from the store if only operationName is defined. There could also be a setting that only allows operationName to be defined and rejects any query with an actual document.

jaulz:
@helfer That approach looks really promising. I will definitely keep an eye on that and wait until it is ready for a hapijs integration :) Until then it seems I should go on with a simple REST API and reuse that later for an GraphQL API.

PR #83, opened 2016-05-01
Prepared queries

calebmer:
Helps people wanting to take `express-graphql` into production by taking a single GraphQL parsed document and only allowing requests to be made against it. Also does not parse and validate queries submitted by the client, such queries should just be a noop. Only `operationName` matters if a parsed document is set. `withDocument` is a working name, any better suggestions?


leebyron:
It's hard to review this since the file has changed - can you maintain the same file so we can review this?

Also cc @alangenfeld may want to take a look

leebyron:
Perhaps a better way to do this would be in a separate middleware? This middleware allows for that by looking to see if `request.body` is already a parsed object.

You could definitely write a middleware which 400 if the user provides `query`, and then otherwise sets:

```js
request.body.query = "your graphql query string here"

calebmer:
That feels like way more work to maintain on our end and the user's end. Is there a reason you'd prefer this functionality in a seperate middleware?

Reverted the file move, thanks for taking a look at these PRs 😊

leebyron:
It doesn't have to be more work. It might be two npm packages instead of one. 

I'm both wary of making this feel like the recommended path when I wouldn't recommend it as a general solution (though fine in some circumstances) and want to avoid a tragedy of the commons situation where this package gets overly complicated with added features.

leebyron:
This is also part of the beauty of the middleware model, it makes extensibility possible

calebmer:
Ok, sounds fine 😊

Let's add support for pre-parsed queries so I could implement this in my own projects.

leebyron:
Closing in favor of the alternative approach

PR #84, opened 2016-05-01
Send response to thunk with request

calebmer:
I use [`on-finished`](https://www.npmjs.com/package/on-finished) to do cleanup at the end of a GraphQL request. Having the response object is necessary for this.


leebyron:
Great idea.

PR #86, opened 2016-05-01
Add documentation for new features

calebmer:
This adds documentation for all my PRs: #82, #83, #84, and #85. Maybe the maintainers would prefer to write the documentation instead?

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,27 +1,33 @@
-GraphQL Express Middleware
-==========================
+GraphQL HTTP Server Middleware
+==============================
 
 [![Build Status](https://travis-ci.org/graphql/express-graphql.svg?branch=master)](https://travis-ci.org/graphql/express-graphql)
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -107,22 +120,72 @@ depending on the provided *Content-Type* header.
     query string, which provides the `query` parameter.
 
 
-### Advanced Options
+## Preparing a Document
+In order to avoid malicious attacks in production, it is recommended to only
+allow approved GraphQL queries to be completely certain none of them will drain
+backend resources in an unintended fashion. Therefore this GraphQL middleware
```
leebyron: Going to remove this in a merge, since that PR is still under discussion. I think this language is too strong, I would personally recommend against this approach for dealing with malicious attacks as it removes one of the primary benefits of GraphQL: that existing clients will not break when servers evolve.

leebyron:
Merged in e96cbf47742de813b083583e46f440f69021464c

PR #93, opened 2016-05-27
Close #92, avoid module.exports breaking change

### LINKED ISSUE #92: 0.5.2 changes module.export
The symbol exported when using commonjs style require has changed.

In 0.5.1:

```
> require('express-graphql')
[Function: graphqlHTTP]
```

In 0.5.2:

```
> require('express-graphql')
{ default: [Function: graphqlHTTP] }
```

Something related to how es modules are transpiled to commonjs I guess.

chentsulin:
We should avoid breaking change introduced by babel6. This add 

```js
module.exports = exports['default'];
```

to the compiled files.

### DIFF
--- a/package.json
+++ b/package.json
@@ -41,6 +41,7 @@
       "es2015"
     ],
     "plugins": [
+      "add-module-exports",
       "transform-class-properties",
       "transform-flow-strip-types"
     ]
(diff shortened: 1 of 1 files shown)


leebyron:
Thanks for catching this.

AlmaasAre:
Unable to require the module in Typescript after this change. It compiles, but seems like the module doesn't have any default property. Which is strange. 


Code:
```
import graphqlHTTP from 'express-graphql';
import schema from './schema';

const graphqlMiddleware:any = graphqlHTTP({
  graphiql: true,
  pretty: true,
  schema: schema,
});

```

runtime error:

```
const graphqlMiddleware = express_graphql_1.default({
                                                                           ^
TypeError: express_graphql_1.default is not a function

```

PR #95, opened 2016-05-29
use res.send directly to send data

chentsulin:
```js
response.write(data);
response.end();
```

can be simplified to 

```js
response.end(data);
```


steelbrain:
This cost us a lot of debugging time 👍

PR #97, opened 2016-06-09
Pass through the logFn option to execute

Slava:
This PR is accommodating https://github.com/graphql/graphql-js/pull/402

Once that PR is merged, this will one will be a logical pass through.

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -101,6 +101,7 @@ export default function graphqlHTTP(options: Options): Middleware {
     let variables;
     let operationName;
     let validationRules;
+    let logFn;
 
     // Promises are used as a mechanism for capturing any thrown errors during
     // the asyncronous process below.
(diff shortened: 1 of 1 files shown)


wincent:
Going to close this out as it is quite stale now. We should revisit this once the details in https://github.com/graphql/graphql-js/pull/402 have been hammered out. Thanks for the PR @Slava!

Slava:
okie :(

Hope we can find a better more elegant solution to this.

wincent:
> Hope we can find a better more elegant solution to this.

Yep, I think the best way to do that is to push on with the discussion in https://github.com/graphql/graphql-js/pull/402 — specifically:

> I'd like to take a step back and propose a handful of alternative logging APIs and have a discussion around the pros/cons of each.

PR #104, opened 2016-07-04
Add option to format result before sending to the client

robrichard:
Adds an option to provide a function to format the result before sending to the client. My use case is to create a debug mode where I can return information about each request our GraphQL server makes to our REST API. As each request is made, the debug information is appended to the context, then returned as part of the json response.

### DIFF
--- a/README.md
+++ b/README.md
@@ -47,6 +47,10 @@ The `graphqlHTTP` function accepts the following options:
     provided, GraphQL's default spec-compliant [`formatError`][] function will
     be used.
 
+  * **`formatResult`**: An optional function which will be used to format the
+    result of the GraphQL operation. Invoked with the result from graphql-js
+    and the context.
+
   * **`validationRules`**: Optional additional validation rules queries must
     satisfy in addition to those defined by the GraphQL spec.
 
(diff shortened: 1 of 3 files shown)


robrichard:
Hey @leebyron, what are your thoughts on this? Is this an appropriate way to accomplish what I'm trying to do or should I use something like https://www.npmjs.com/package/express-mung instead?

natew:
I'd actually like something similar but *before* returning result. As of now there's no way to add authentication or caching without a hook into the graphql response part. I think a simple option for hooking into the request before and after, or letting us return an async function that does our own request.

helfer:
@natew Just in case it's of interest to you, we've implemented hooks for exactly that in [Apollo Server](https://github.com/apollostack/apollo-server). You can modify the query, variables etc. before executing, and you can process the result before sending it back to the client.

Under the hood, Apollo Server uses graphql-js just like express-graphql.

leebyron:
Sorry for the slow response getting to this. We explicitly do not want to support this as it breaks GraphQL spec compliance. For more on this see: https://github.com/graphql/express-graphql/pull/118#issuecomment-258027222

I'll be adding a similar mechanism for adding additional metadata to responses that is spec compliant for the next version!

PR #106, opened 2016-07-19
return undefined instead of null in safeSerialize to make localStorage functionality back

jarvisaoieong:
According to the implementation in [graphiql](https://github.com/graphql/graphiql/blob/master/src/components/GraphiQL.js#L139) , I think we should change `null` to `undefined` in order to enable the localStorage functionality in graphiql.


facebook-github-bot-1:

leebyron:
Great suggestion.

leebyron:
Had to merge this manually

PR #109, opened 2016-07-20
[RFC] Persisted Document support

alangenfeld:
This adds support to express-graphql for GraphQL persisted documents. This is a feature we have discussed in ad-hoc ways since open sourcing GraphQL, and are now attempting to demonstrate more concretely. Persisted documents are foundational to how we use GraphQL at Facebook, especially for our mobile clients. Additional background context is being added to the README.

The goal in adding this support to express-graphql is to provide a clear example of how this idea can be implemented which should allow it to be translated to any other GraphQL server middleware or implementation.

Notes:
- persistValidatedDocument and providing the /persist endpoint is still a nascent idea, that needs the most scrutiny. Then once its settled, tests.

### DIFF
--- a/README.md
+++ b/README.md
@@ -56,6 +56,15 @@ The `graphqlHTTP` function accepts the following options:
   * **`validationRules`**: Optional additional validation rules queries must
     satisfy in addition to those defined by the GraphQL spec.
 
+  * **`loadPersistedDocument`**: A function that takes an input id and returns a
+    valid Document. If provided, this will allow your GraphQL endpoint to execute
+    a document specified via `documentID`.
+
+  * **`persistValidatedDocument`**: A function that takes a validated Document and
+    returns an id that can be used to load it later. This is used in conjunction
+    with `loadPersistedDocument`. Providing this function will enable persisting
+    documents via the `document` parameter at the `/persist` subpath. It is
+    recommended that this option only be enabled in development deployments.
 
 ## HTTP Usage
 
@@ -140,6 +149,21 @@ new GraphQLObjectType({
 });
 ```
 
+## Persisted Documents
+
+Putting control of the query in to clients hands is one of the primary benefits
+of GraphQL. Though when this is deployed to production naively, it is possible to
+end up uploading and revalidating the same document text across all clients of a
+particular deployment. It can even be problematic within a single client, sending
+the exact same query repeatedly. Uploading and validating a large GraphQL document
+can be very costly, especially over a poor network connection from a mobile device
+with something large and complex such as Facebook's News Feed GraphQL query.
+
+One solution to this problem that was developed early in the use of GraphQL at
+Facebook was persisted documents. At build time, we take the GraphQL document and
+persist it on the server getting back an ID. Then at runtime, we only upload the
+ID of the document and the variables. When the server receives this request, it
+loads the query and executes it, knowing it was already validated.
 
 ## Debugging Tips
 
@@ -153,7 +177,6 @@ formatError: error => ({
   stack: error.stack
 })
 ```
-
 
 [`GraphQL.js`]: https://github.com/graphql/graphql-js
 [`formatError`]: https://github.com/graphql/graphql-js/blob/master/src/error/formatError.js
--- a/README.md
+++ b/README.md
@@ -151,7 +151,7 @@ new GraphQLObjectType({
 
 ## Persisted Documents
 
-Putting control of the query in to clients hands is one of the primary benefits
+Putting control of the query into clients' hands is one of the primary benefits
 of GraphQL. Though when this is deployed to production naively, it is possible to
 end up uploading and revalidating the same document text across all clients of a
 particular deployment. It can even be problematic within a single client, sending
--- a/src/__tests__/http-test.js
+++ b/src/__tests__/http-test.js
@@ -28,7 +28,8 @@ import {
   GraphQLNonNull,
   GraphQLString,
   GraphQLError,
-  BREAK
+  BREAK,
+  parse
 } from 'graphql';
 import graphqlHTTP from '../';
 
@@ -76,6 +77,14 @@ const TestSchema = new GraphQLSchema({
 
 function urlString(urlParams?: ?Object) {
   let string = '/graphql';
+  if (urlParams) {
+    string += ('?' + stringify(urlParams));
+  }
+  return string;
+}
+
+function persistString(urlParams?: ?Object) {
+  let string = '/graphql/persist';
   if (urlParams) {
     string += ('?' + stringify(urlParams));
   }
@@ -212,6 +221,72 @@ describe('test harness', () => {
         });
       });
 
+      it('allows GET with documentID param', async () => {
+        const app = server();
+
+        app.use(urlString(), graphqlHTTP({
+          schema: TestSchema,
+          loadPersistedDocument: async () => parse('{test}')
+        }));
+
+        const response = await request(app)
+          .get(urlString({
+            documentID: '1'
+          }));
+
+        expect(response.text).to.equal(
+          '{"data":{"test":"Hello World"}}'
+        );
+      });
+
+      it('allows GET with documentID and variable values', async () => {
+        const app = server();
+
+        app.use(urlString(), graphqlHTTP({
+          schema: TestSchema,
+          loadPersistedDocument: async() => parse('query helloWho($who: String){ test(who: $who) }')
+        }));
+
+        const response = await request(app)
+          .get(urlString({
+            documentID: '1',
+            variables: JSON.stringify({ who: 'Dolly' })
+          }));
+
+        expect(response.text).to.equal(
+          '{"data":{"test":"Hello Dolly"}}'
+        );
+      });
+
+      it('allows GET with documentID and operation name', async () => {
+        const app = server();
+
+        app.use(urlString(), graphqlHTTP(() => ({
+          schema: TestSchema,
+          loadPersistedDocument: async() => parse(`
+              query helloYou { test(who: "You"), ...shared }
+              query helloWorld { test(who: "World"), ...shared }
+              query helloDolly { test(who: "Dolly"), ...shared }
+              fragment shared on QueryRoot {
+                shared: test(who: "Everyone")
+              }
+            `)
+        })));
+
+        const response = await request(app)
+          .get(urlString({
+            documentID: '1',
+            operationName: 'helloWorld'
+          }));
+
+        expect(JSON.parse(response.text)).to.deep.equal({
+          data: {
+            test: 'Hello World',
+            shared: 'Hello Everyone',
+          }
+        });
+      });
+
       it('Reports validation errors', async () => {
         const app = server();
 
@@ -325,6 +400,31 @@ describe('test harness', () => {
           data: {
             test: 'Hello World'
           }
+        });
+      });
+
+      it('Errors when sending a mutation via GET by documentID', async () => {
+        const app = server();
+
+        app.use(urlString(), graphqlHTTP({
+          schema: TestSchema,
+          loadPersistedDocument: async() => parse(`
+            mutation TestMutation { writeTest { test } }
+          `)
+        }));
+
+        const error = await catchError(
+          request(app)
+            .get(urlString({
+              documentID: '1'
+            }))
+        );
+
+        expect(error.response.status).to.equal(405);
+        expect(JSON.parse(error.response.text)).to.deep.equal({
+          errors: [
+            { message: 'Can only perform a mutation operation from a POST request.' }
+          ]
         });
       });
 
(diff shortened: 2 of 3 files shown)


alangenfeld:
I am not the greatest JS engineer, so basic/style feedback is expected and welcome.

src/index.js
```diff
@@ -297,3 +304,127 @@ function canDisplayGraphiQL(
...
+
+/**
+ * Helper function to resolve the Options to get OptionsData.
+ */
+function resolveOptionsData(
+  options: Options,
+  request: Request,
+  response: Response
+): Promise<OptionsData> {
+  return new Promise(resolve => {
```
kassens: No need for `new Promise`, this should work too:

```
return Promise.resolve(
   typeof options === 'function' ? options(request, response) : options
).then
```

src/index.js
```diff
@@ -226,6 +216,11 @@ export default function graphqlHTTP(options: Options): Middleware {
         return { errors: [ contextError ] };
       }
     }).catch(error => {
```
alangenfeld: [1]

src/index.js
alangenfeld: tried that, but doing so prevents us from catching exceptions thrown from options when it is a function at [1]

nodkz:
@alangenfeld @wincent @josephsavona why `persisted document`?

Document can't have variables for fetching. According to provided variables, we may get the absolutely different responses for same document (query) id.

I think better naming is `persisted query`. Word `document` brings to much ambiguity.

dschafer:
Yeah, this is a tricky nomenclature issue. The parameter is `params.query`, but `params.query` doesn't have to contain a GraphQL Query; it has to contain a GraphQL Document. So I could see either name working here.

I'm not sure Document is ambiguous in this context, since it refers to http://facebook.github.io/graphql/#sec-Language.Query-Document (or to the `Document` flowtype in graphql-js), but it doesn't match the existing parameter naming.

nodkz:
`QueryDocument` will be a nice name, but not just `document`. Anyway for me, `PersistedQuery` is the best naming.

@dschafer please consider better name for flowtype parameter. `Document` is not good.

README.md
```diff
@@ -53,6 +53,16 @@ The `graphqlHTTP` function accepts the following options:
...
 
+  * **`loadPersistedDocument`**: A function that takes an input id and returns a
+    valid Document. If provided, this will allow your GraphQL endpoint to execute
+    a document specified via `documentID`.
+
+  * **`persistValidatedDocument`**: A function that takes a validated Document and
+    returns an id that can be used to load it later. This is used in conjunction
+    with `loadPersistedDocument`. Providing this function will enable persisting
+    documents via the `document` parameter at the `/persist` subpath. It is
+    recommended that this option only be enabled in development deployments.
```
lacker: It's not clear to me when you would want to use `persistValidatedDocument` in development. You basically have to be storing the persisted documents either in some database, or in your source code, right? And either way, hitting a `persistValidatedDocument` REST API on a development backend server is not going to be the easiest way to get these persisted documents into the place they need to go. In your development environment you can just write them to the db or put them in your source code directly.

README.md
```diff
@@ -53,6 +53,16 @@ The `graphqlHTTP` function accepts the following options:
   * **`graphiql`**: If `true`, may present [GraphiQL][] when loaded directly
     from a browser (a useful tool for debugging and exploration).
 
+  * **`loadPersistedDocument`**: A function that takes an input id and returns a
+    valid Document. If provided, this will allow your GraphQL endpoint to execute
+    a document specified via `documentID`.
```
lacker: These documents typically each have a unique name, right? Why not use the name as the identifier instead of a separate documentID? I think that is how Airbnb is setting up their graphql backend.

README.md
```diff
@@ -156,6 +166,22 @@ new GraphQLObjectType({
 });
 ```
 
+## Persisted Documents
+
+Putting control of the query in to clients hands is one of the primary benefits
```
lacker: in to -> into
clients -> clients'

josephsavona:
@lacker the problem with using the document name as the identifier is that the document changes over time, the id serves as a hash for a rev of a particular document.

README.md
kassens: While the document might have a unique name like `HomePageQuery`, this is not enough, as the query likely changes over time as new client versions are released. It is important that old clients still get the old version of `HomePageQuery`.

README.md
kassens: If you want to build tooling for client development, you might want to include persisting all required client queries into your build process and have an API for this, making this interface HTTP is pretty obvious.

Persisting them to a database has the nice benefit (especially for larger organizations) that you don't need a new server deployment after adding (updating) a new query. You could even imagine that Github's public GraphQL API allowed clients to persist queries.

alangenfeld:
Apologies for the long delay, its been a hectic few months in GraphQL land at FB. Thanks everybody for the feedback so far. I think the questions that are still open are:

- naming: persisted document vs. persisted query
- whether supporting `/graphql/persist?document=`  if you provide persistValidatedDocument makes sense

README.md
lacker: You would have to be running your backend server during your build process, though - isn't that somewhat unusual? Rather than writing some backend code to persist the document, and making your build process hit an HTTP API on your running backend server, your build process could just execute the code to persist the document directly. I guess overall it seems like this is something people could implement themselves just as easily as if we added a `persistValidatedDocument` API.

lacker:
My suggestions + rationale for the open questions:

* "persisted query" because in the graphql community more people refer to them as "queries" than as "documents"
* no support for /graphql/persist?document=. It doesn't seem like this provides enough benefit over people implementing their own endpoint.

helfer:
In apollo-server we called them "stored operations", since that encompasses queries, mutations and subscriptions. I know that technically they can also contain fragments, which makes them documents, but at least it's a less confusing name than "queries". Using "queries" may lead people to believe that mutations can't be persisted.

README.md
leebyron: I don't think it's unusual at all - in fact it seems the opposite to me. When building my iOS app or Android app or web app, I'd be really surprised if in that context I couldn't access my GraphQL API because it was also being built. I would expect a GraphQL API to be decoupled from the apps that are using it such that the build and deploy processes are also separate. Typically the build process for an iOS, Android or web app does not have access to a database, which is why exposing an HTTP API is beneficial.

leebyron:
Let's revive this and get it shored up and ready to land.

Some thoughts:

"Document" is absolutely the right name. We've been bad about clarity around names and in fact should consider renaming some of the fields accepted in this library to help with that. A Document is a whole GraphQL file or request. An operation is one query/mutation (a document may contain many in addition to fragments). A query is a kind of operation. Since we're typically talking about documents that have one operation that is a query, we say "GraphQL query" - but I think it's important to get the naming right for this which should allow persisting any kind of document.

Perhaps we should rearrange the API such that you provide both or neither functions. Right now we could assert one or the other, but I think the API could help with that. Maybe:

```js
persistedDocuments: {
  save: document => id 
  load: id => document
}
```

I think we're missing an operation as well, which is getting the actual document text for a given id. Tools like GraphiQL would want to use this to populate the left pane. Maybe we should expose paths `savepersisted` and `loadpersisted` to represent both?

One last thought - should we be persisting the actual string text rather than the AST itself? Or perhaps allowing either with a `typeof === 'string'` check? I'm thinking if your document contains comments or preferred formatting, that a round trip through persistence should retain those.

josh:
> should we be persisting the actual string text rather than the AST itself?

I've been curious about the same thing for @github. Right now we're thinking of making our persisted IDs a SHA256 of the document text. Normalizing the document AST would let us store slightly different text documents under the same hashed ID. Seems like a good idea to me. Is there even a mechanize that would allow clients to retrieve the original document text? If they can only persist and execute operations, any comments wouldn't make a difference.

leebyron:
I think it would be valuable to retrieve the original document text for use with tools like GraphiQL. It would be immensely helpful for debugging to be able to take one of these persisted IDs and check what the server is actually running for it.

leebyron:
Also, if you'd prefer to have comments be irrelevant to the hash of a query, that's certainly something you could have a client decide by doing a `hash(print(parse(text)))` to normalize first.

JeffRMoore:
Is there some value here to borrowing the prepared statement nomenclature from SQL?  Many programmers are already familiar the concept of working with a statement in two phases, one phase to setup the statement and the second phase to execute it with parameters.

The methods in this example might become prepareDocument and loadPreparedDocument?

I wonder if "persist" terminology might require slightly more cognition to disambiguate. It is the communication pattern with the server that is being persisted, not necessarily the data being exchanged.

src/index.js
```diff
@@ -295,3 +302,126 @@ function canDisplayGraphiQL(
...
+  request: Request,
+  response: Response
+): Promise<void> {
+  let pretty;
+  let persistValidatedDocument;
+  return Promise.all([
+    resolveOptionsData(options, request, response),
+    parseBody(request)
+  ]).then(initData => {
+    const [ optionsData, bodyData ] = initData;
```
intellix: travis.yml contains Node 0.12 with all tests currently passing. I believe destructuring isn't supported there.

leebyron:
I'll take over from here :)

alangenfeld:
Thanks @leebyron! Apologies for not being able to see this one through.

raderio:
Is someone using approach: on dev environment support mix(persisted queries & arbitrary queries) and on prod environment support only persisted queries?

taion:
What's the status on this PR?

j0k3r:
Any news?

alloy:
@leebyron @kassens We’re now starting to use persisted queries at Artsy and so I’d like to contribute some time to take this PR over the finish line. Could you summarise the work left to do from your perspective?

src/__tests__/http-test.js
```diff
@@ -1498,5 +1598,47 @@ describe('test harness', () => {
...
+        app.use(urlString(), graphqlHTTP({
+          schema: TestSchema,
+          persistValidatedDocument: persist,
+          loadPersistedDocument: load
+        }));
+
+        let response = await request(app)
+          .post(persistString({
+            document: '{test}'
+          }));
```
leebyron: Need a test that posting data works instead of just formatting a query string

src/index.js
```diff
@@ -158,35 +155,28 @@ export default function graphqlHTTP(options: Options): Middleware {
...
 
       // If there is no query, but GraphiQL will be displayed, do not produce
       // a result, otherwise return a 400: Bad Request.
-      if (!query) {
+      if (!query && !documentID) {
         if (showGraphiQL) {
           return null;
         }
-        throw httpError(400, 'Must provide query string.');
+        throw httpError(400, 'Must provide query string or document ID.');
```
leebyron: Should probably only say document ID should be provided if `loadPersistedDocument` is available

src/index.js
```diff
@@ -158,35 +155,28 @@ export default function graphqlHTTP(options: Options): Middleware {
...
+        return loadPersistedDocument(documentID);
       }
-
-      // Validate AST, reporting any errors.
-      const validationErrors = validate(schema, documentAST, validationRules);
-      if (validationErrors.length > 0) {
-        // Return 400: Bad Request if any validation errors exist.
-        response.statusCode = 400;
-        return { errors: validationErrors };
+      if (query) {
```
leebyron: These 3 if statements are a bit confusing in sequence, I would expect:

```
if (documentID) {
} else if (query) {
} else {
```

src/index.js
```diff
@@ -278,7 +285,7 @@ function getGraphQLParams(urlData: Object, bodyData: Object): GraphQLParams {
   // Name of GraphQL operation to execute.
   const operationName = urlData.operationName || bodyData.operationName;
 
-  return { query, variables, operationName };
+  return { query, variables, operationName, documentID};
```
leebyron: formatting (note: we should enable prettier on this repo)

alloy:
@leebyron Looks straightforward enough, thanks! 👍 I’ll start work on that this week.

sibelius:
let's make this happen in 2019?

alloy:
I made many attempts at updating this PR, but the lack of context and it being so out of date made it hard :/ It’s definitely still on my TODO list, so yeah maybe this will be the year!

sibelius:
this could be achieve in a middleware based way, so we don't need to make express-graphql more complex

like this https://gist.github.com/robrichard/d5a6ae7ecb3c9efc7fad7b3236d255e7 solution by @robrichard 

using the middleware approach makes easy to work with express-graphql, koa-graphql, apollo-server and any other middleware based server

josephsavona:
@sibelius that seems like a great approach

Janpot:
But it should be standardized, no? To guarantee interoperability between the different client and server libraries. Middleware or not is just an implementation detail

alloy:
Yeah agreed that that could be a good approach 👍 I do think we may still need to add some change here (eg to be able to use already parsed and validated queries), but those should be easy to do from scratch.

sibelius:
we could have another package to handle all this persisted queries validation and logic

PR #117, opened 2016-08-25
Context defaulting to request

lacker:
This PR slightly tweaks the default behavior so that by default the express request is passed as the context object, and updates a decent amount of the README documentation.

The rationale is to make it simpler to support express middleware, in particular authentication. Most express middleware works similarly - some extra field is tacked onto `request`. This is convenient for express developers because the `request` object is accessible from every route handler. The graphql equivalent of a route handler is a resolver, so if the `request` object is simply available to every resolver, express middleware is generally usable without having to alter the graphqlHTTP route handler itself. I think this is an "expressy" thing to do, even if it does not quite aesthetically jive with the deep love of immutability reflected in graphql itself.

Middleware that behaves this way & can now be used without altering the graphqlHTTP method call:
* express-session tacks on request.session
* express-jwt tacks on request.user
* passport tacks on request.user
* multer tacks on request.file or request.files

Note the "advanced scenario" docs of authentication are simpler now - you do not have to pass a function to the graphqlHTTP method, you can just use the same call that was in the "intro".

This is backward compatible unless you wrote code that relies on having a falsy context which seems unlikely.

Documentation changes are mostly self-explanatory. The change to using port 4000 in the example is so that it doesn't conflict by default with create-react-app - a small tweak but this may avoid little bits of confusion. Using 'require' instead of 'import' is a bit sad but preferable for now because it means that this code works with node out of the box - most node developers are not transpiling.

### DIFF
--- a/README.md
+++ b/README.md
@@ -4,7 +4,7 @@ GraphQL HTTP Server Middleware
 [![Build Status](https://travis-ci.org/graphql/express-graphql.svg?branch=master)](https://travis-ci.org/graphql/express-graphql)
 [![Coverage Status](https://coveralls.io/repos/graphql/express-graphql/badge.svg?branch=master&service=github)](https://coveralls.io/github/graphql/express-graphql?branch=master)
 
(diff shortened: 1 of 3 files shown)


leebyron:
Love it. Great docs improvements as well

PR #119, opened 2016-09-02
Re-enable express etag support

### LINKED ISSUE #115: Broken express etag support
Hi,

The commit 48e6659945477bd706fe22173f450c16eddd5bf5 broke the built-in etag support of express. 

This is due to using `response.write` instead of `response.send` (which has the etag generation inside) https://github.com/graphql/express-graphql/blob/master/src/index.js#L250.

The commit states that this is made to be compatible with connect. Is there a way to also support etags for express?

Thanks,
Martin

brandondoran:
The express `response.send` provides some extra functionality, such as adding an ETag header. https://github.com/graphql/express-graphql/commit/48e6659945477bd706fe22173f450c16eddd5bf5 removed `.send` in favor of `.end` in order to support connect. This PR adds a helper function, `sendResponse`, which calls `response.send` if that method exists (i.e. the server is express), otherwise fallback to `response.end`.

In order to test that the correct method is called, I added sinon as a devDependency and a single test that uses a spy to verify that `send` is called when the server is express. 

This fixes #115

### DIFF
--- a/package.json
+++ b/package.json
@@ -91,6 +91,7 @@
     "multer": "1.1.0",
     "raw-body": "2.1.6",
     "sane": "1.3.4",
+    "sinon": "^1.17.5",
     "supertest": "1.0.1",
     "supertest-as-promised": "2.0.2"
   },
--- a/src/__tests__/http-test.js
+++ b/src/__tests__/http-test.js
@@ -13,6 +13,7 @@
 
 import { expect } from 'chai';
 import { describe, it } from 'mocha';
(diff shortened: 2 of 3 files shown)


facebook-github-bot-0:

facebook-github-bot-9:

vitosamson:
+1 for this change. `.end()` doesn't seem to set the `Content-Length` header which is causing issues for me.

leebyron:
Thanks for this nice detail!

PR #120, opened 2016-09-08
Small tweaks to the README

lacker:
1/ Consistently using "GraphQL.js" rather than "graphql-js"
2/ Kinda crazy but for the longest time I had been mistakenly thinking that you had to navigate to `/graphql/graphiql` to get to GraphiQL by default. Nope, it displays whenever text/html is accepted. BUT - that works anyway accidentally because it's mounted as middleware rather than an endpoint. Anyway I just removed the phrasing that alluded to this past confusion of mine.

### DIFF
--- a/README.md
+++ b/README.md
@@ -32,19 +32,19 @@ app.listen(4000);
 
 The `graphqlHTTP` function accepts the following options:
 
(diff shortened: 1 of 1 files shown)


leebyron:
👌

PR #121, opened 2016-09-16
Export parseBody

stephenbaldwin:
`parseBody` handles complexities in deriving the params from POST requests. It should be exposed to allow for applications and middleware to have access to params like "query" and "operationName" without having to reimplement the logic already defined in `parseBody`.

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -26,6 +26,7 @@ import { renderGraphiQL } from './renderGraphiQL';
 
 import type { Request, Response } from 'express';
 
+export { parseBody };
 
 /**
  * Used to configure the graphqlHTTP middleware by providing a schema


facebook-github-bot-0:

facebook-github-bot-4:

lacker:
IMO this seems useful - you do want middleware to have access to things like the query and the operation name for logging. It probably needs some documentation & tests, though, because if we're exposing it, it's important we don't break it. Could you add some?

stephenbaldwin:
Sure thing @lacker

leebyron:
I think this is a good idea, but I think this particular function is a bit low-level since it doesn't do anything GraphQL specific per-se, instead what you might be looking for is a function that given a request gives you back the GraphQL request details.

I can prep a change that adds this because I agree it would be useful.

PR #133, opened 2016-10-24
update flow to 0.33

chentsulin:
Explicitly return string in `safeSerialize` to avoid `null` cannot be coerced to `string` error


leebyron:
Merged manually, thanks!

PR #153, opened 2016-11-07
Custom context function with request object as an argument

mnmtanish:
Currently, we can only use either the request object or a custom context object. By accepting a function for context, we can use both at the same time.

Note: `npm install` fails so I couldn't run tests

### DIFF
--- a/src/__tests__/http-test.js
+++ b/src/__tests__/http-test.js
@@ -381,6 +381,36 @@ describe('test harness', () => {
         });
       });
 
+      it('Allows passing in a context function', async () => {
+        const app = server();
+
+        // Middleware that adds req.foo to every request
+        app.use((req, res, next) => {
+          req.foo = 'bar';
+          next();
+        });
+
+        app.use(urlString(), graphqlHTTP({
+          schema: TestSchema,
+          context: req => `dotFoo:${req.foo}`
+        }));
+
+        const response = await request(app)
+          .get(urlString({
+            operationName: 'TestQuery',
+            query: `
+              query TestQuery { context }
+            `
+          }));
+
+        expect(response.status).to.equal(200);
+        expect(JSON.parse(response.text)).to.deep.equal({
+          data: {
+            context: 'dotFoo:bar'
+          }
+        });
+      });
+
       it('Allows returning an options Promise', async () => {
         const app = server();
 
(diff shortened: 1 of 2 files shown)


nodkz:
Suppose that @leebyron [YAGNI](https://ru.wikipedia.org/wiki/YAGNI)sh (reject) this PR.

Current implementation (context as object):
```js
import graphqlHTTP from 'express-graphql';
server.use('/graphql', (req) => ({
  schema: GraphQLSchema,
  context: {
    userId: req.userId,
  },
}));
```

Current implementation (context as function):
```js
import graphqlHTTP from 'express-graphql';
function prepareMyAwesomeContext(req) {
  return { userId: req.userId };
}
server.use('/graphql', (req) => ({
  schema: GraphQLSchema,
  context: prepareMyAwesomeContext(req),
}));
```

Your implementation, just add syntax sugar:
```js
import graphqlHTTP from 'express-graphql';
function prepareMyAwesomeContext(req) {
  return { userId: req.userId };
}
server.use('/graphql', (req) => ({
  schema: GraphQLSchema,
  context: prepareMyAwesomeContext, // <-------- Just remove `(req)`
}));
```

mnmtanish:
Hi @nodkz 

I didn't quite get the example code. Can you please fix them? It'll be a great help for me. Thanks.

***

I know I can do it like this:

```js
import graphqlHTTP from 'express-graphql';
server.use('/graphql', (req, res, next) => {
  req.foo = 'bar';
  next();
});
server.use('/graphql', graphqlHTTP({
  schema: GraphQLSchema,
}));
```

But if the app code and other express middleware all use the same request object like this, it can easily cause unexpected errors later.

My second option is to skip `express-graphql` and use the `graphql-js` directly. Which is what I'm planning to do if this didn't get accepted.

Supporting functions for context is useful in more ways than this. We can prepare resources we may need to process a request independently for each request. So, one request cannot change the context object and mess with other requests.

nodkz:
Change
```js
server.use('/graphql', graphqlHTTP({
  schema: GraphQLSchema,
}));
```
by
```js
server.use('/graphql', graphqlHTTP((req) => ({ // will be called individually for every request
  schema: GraphQLSchema,
})));
```
Don't worry about the mess and unexpected errors later. Re-read how express middlewares works, if so.

mnmtanish:
Thanks @nodkz

PR #157, opened 2016-11-13
Removed content-type from devDependencies

chentsulin:
It already be listed in `dependencies`


wincent:
Thanks, @chentsulin. Good catch.

PR #158, opened 2016-11-13
added Response to option thunk type second argument

chentsulin:


wincent:
Looks reasonable on the surface, @chentsulin, but can you share a motivating use case for requiring the response here?

chentsulin:
If I’m not mistaken, response already passed as second argument to option: https://github.com/graphql/express-graphql/blob/master/src/index.js#L164-Lundefined

wincent:
Looks like you're right. Thanks for clarifying!

PR #159, opened 2016-11-14
use 'const' and 'let' in resources/watch.js

chentsulin:


wincent:
Thanks, @chentsulin! Sorry for the delay in getting to this.

PR #184, opened 2017-01-12
Increased hard limit for content body to 1mb

juniorplenty:
Also committed build directory, for easy install via git


juniorplenty:
Oops, accident

PR #199, opened 2017-01-25
Drop 0.12 in CI

chentsulin:
node 0.12 is unmaintained


wincent:
Seems reasonable to me, and will make CI runs a little faster. It will obviously still work in 0.12 for those who care to run it there, but I don't think we need to give it "official support". Thanks for the PR.

PR #218, opened 2017-02-14
Pass request to extensionFn

jfizz:
in my project, i remove fields in the extension function based on the requesting user. since it is common to store the user on the request object, i thought this might be useful for others. i do not remove the fields in the resolve function so that child-resolves have access to a fully-hydrated parent resource.

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -284,7 +284,8 @@ function graphqlHTTP(options: Options): Middleware {
           document: documentAST,
           variables,
           operationName,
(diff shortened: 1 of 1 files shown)


jfizz:
On second thought, this is the wrong approach. The resource maybe unidentifiable when passing through the extension function depending on the selected fields. Sorry for the noise.

wincent:
Thanks for closing the loop!

Issue #112, opened 2016-08-10
Allow fallback to hosted js files without cdn


ofbeaton:
Tried to use graphiql provided by this package on an intranet without an internet connection, horrible experience, everything is loaded from cdn.

Similar issues in other flavours of graphiql, seems like I'm not the only one having this issue.
exponentjs/koa-graphiql#1

This was already fixed in the `example/` for graphiql itself: graphql/graphiql#135

Could we please have a fallback where if the cdn fails to load the files it loads a local copy?

leebyron:
Good idea. Should be able to pass something like "local" or "cdn" to the graphiql option

This would be a good pick up task for anyone interested in contributing.

ofbeaton:
or perhaps something like bellow? Though the advantage of the local flag is that it wouldn't even bother trying to hit the external network.

```
<script src="//ajax.googleapis.com/ajax/libs/jquery/1.11.3/jquery.min.js"></script>
<script>window.jQuery || document.write('<script src="js/vendor/jquery.min.js') }}"><\/script>')</script>
```

ilia-luk:
+1

KyleAMathews:
Or just make GraphiQL a react app and compile it out to a bundle.js.

dgrcode:
I'm happy to work on this if nobody is already doing it

tanaypratap:
great. I will be working on this and putting up a draft peer soon. Thanks for the quick reply @dgrcode

IvanGoncharov:
CDN usage was completely removed in #524

PR #219, opened 2017-02-15
Add option to load graphiql files locally

CorinChappy:
An implementation of https://github.com/graphql/express-graphql/issues/112, allows a user to pass "local" or "cdn" to the graphiql option in order to specify where the files should be loaded from. Is also compatible with true/false, defaulting to "cdn" if true.

I put the files in a new folder called vendor, but I am happy to change it to work a different way (using the files from npm for example), if that would be better.

I've not used flowtype before so I hope I have covered everything there! And no tests or documentation changes yet, but I can also look into adding them!

Note: I have had to change what is returned by the graphqlHTTP function, it now returns an express router instead of just a middleware function, this is so the vendor files can be served using express.static under the same route as the graphql endpoint


jacobbogers:
any progress on this PR?

gilesbradshaw:
It would be really good to be able to do this

IvanGoncharov:
Closed by #524 that fully removes CDN usage.

PR #240, opened 2017-03-06
Invalid JSON error when not passing any value in Query Variable of GraphiQL

rishirajgarg:
Query Variables should not be required, but they must be valid JSON when they are required.

In the current scenario, it's throwing Invalid JSON request when there is a blank value in Query Variable.


src/index.js
```diff
@@ -362,7 +362,7 @@ function parseGraphQLParams(
 
   // Parse the variables if needed.
   let variables = urlData.variables || bodyData.variables;
-  if (typeof variables === 'string') {
+  if (variables !== null && variables !== '' && typeof variables === 'string') {
```
asiandrummer: Let's just check for falsyness:
```
if (variables && typeof variables === 'string')
```

rishirajgarg: thanks @asiandrummer - incorporated and tested the changes. looks good so far.

 if (variables && typeof variables === 'string') {
   ...
 }

Issue #197, opened 2017-01-24
Raw GraphQL request on req object


richardwillars:
Hello

I'd like to output the original GraphQL query (and any variables) to the console whenever an error is thrown in order to make debugging easier. At the moment I can't see any way of doing this as req.body isn't available as express-graphql is doing it's own body parsing, and the raw query doesn't seem to get passed anywhere.

The closest thing I've found so far is fieldNodes which is an object describing the query, but it's still quite difficult to reverse engineer the original query from this.

Could we get the raw query added to the express req object?

Thanks!

nodkz:
Maybe this issue will be helpful for you https://github.com/graphql/graphql-js/issues/662
and this [custom error catching realization]( https://gist.github.com/nodkz/d14b236d67251d2df5674cb446843732)

Also `graphql-express` has `extensions` read [here](https://github.com/graphql/express-graphql#providing-extensions). It may be a function, which contains `document` argument. If you need a raw `query` try to open PR with adding query to args [here](https://github.com/graphql/express-graphql/blob/c92cfad64f4b18abce3cdeca36bb8553dd020773/src/index.js#L283) (query is populated in upper scope, so it is easy to add just dropping a one line `query,`)

asiandrummer:
@richardwillars Using one of the arguments (`document`), shouldn't you be able to do something like: `print(document)` and get the query string?

richardwillars:
If I was using GET I guess that'd work, though in my case I'm using POST. Don't think there's anything in there for that

shtse8:
any updates?

robrichard:
I think you can do this if you add your own body parser.

```js
const graphqlHTTP = require('express-graphql');
const bodyParser = require('body-parser');
const {formatError} = require('graphql');

app.use(bodyParser.json()); // support json encoded bodies
app.use(bodyParser.urlencoded({ extended: true })); // support encoded bodies

app.use('/', graphqlHTTP(function (req) {
    return {
        schema: Schema,
        graphiql: true,
        rootValue: {},
        context: buildContext(req),
        formatError: e => {
               const query = req.body.query;
               const variables = req.body.variables;
               console.log('Error!', query, JSON.stringify(variables, null, 2));
               return formatError(e);
       }
    };
}));

```

PR #248, opened 2017-03-14
Extensions function receives raw query param

eladitzhakian:
Following @nodkz 's suggestion [here](https://github.com/graphql/express-graphql/issues/197#issuecomment-274791235) (and also, I really need this 😄 )

### DIFF
--- a/README.md
+++ b/README.md
@@ -58,7 +58,7 @@ The `graphqlHTTP` function accepts the following options:
     `"extensions"` field in the resulting JSON. This is often a useful place to
     add development time metadata such as the runtime of a query or the amount
     of resources consumed. This may be an async function. The function is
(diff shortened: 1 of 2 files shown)


asiandrummer:
@eladitzhakian we should able to get the query string by printing the document AST (`print(ast)`) - would that suffice your use case?

eladitzhakian:
@asiandrummer It would work, and it's not the only workaround as the query string is also available via the `req` variable, but I still believe it belongs in a dedicated arg

leebyron:
Hey @eladitzhakian, sorry for the wait. #302 now ensures that this value is available for all options including extensions - however you will need to refer to the value provided in the options function - does that work for you?

eladitzhakian:
@leebyron sure, thanks

Issue #216, opened 2017-02-14
Use for Restify requires not stringifying json


terencechow:
I'm not sure what [this line](https://github.com/graphql/express-graphql/blob/master/src/index.js#L323) is doing but it makes this repo not compatible with Restify since the response is returned as a stringified JSON. Note that in Express 4.x you don't need to stringify the JSON either. 

For reference here the line that is causing problems is 

`var_payload = JSON.stringify(result, null, pretty? 2:0)`: 

in the follow code:
```
 // Otherwise, present JSON directly.
        var _payload = JSON.stringify(result, null, pretty ? 2 : 0);
        response.setHeader('Content-Type', 'application/json; charset=utf-8');
        sendResponse(response, result);
```

Is this line for compatibility of an older Express? If not can we remove it so that this repo can be used with Restify? I can make a PR if you want.

asiandrummer:
> Is this line for compatibility of an older Express?

Not sure what the exact intention at the time was, but since we're @ Express 4.x anyways I think we can go ahead and do that. cc @leebyron though.

santino:
Any update on this?
I'm facing the same issue using express-graphql with restify.

Basically this block of code
```
  // Otherwise, present JSON directly.
  const payload = JSON.stringify(result, null, pretty ? 2 : 0);
  response.setHeader('Content-Type', 'application/json; charset=utf-8');
  sendResponse(response, payload);
```

Could just be simplified into this:
```
  // const payload = JSON.stringify(result, null, pretty ? 2 : 0);
  response.setHeader('Content-Type', 'application/json; charset=utf-8');
  sendResponse(response, result);
```

This is causing the response to be stringified twice which is breaking the implementation with a Relay client app.

It would be good to know what the plan is to tackle this issue.

wincent:
I think we'd accept a PR on this. If this requires Express >= 4 then we might want to add that as a peer dependency. Currently it's only a dev dependency.

santino:
I'm looking into making a PR for this.
It's not as straightforward as it might seem if you want to stay compatible with all your current features.

First of all if we delegate responsibility of stringifying JSON to the server you will lose the 'pretty'  printing option.
Also compatibility with Connect might be an issue as I don't believe they stringify the response by default like express and restify, and they don't have a json function on their response method (`res.json({foo: 'bar'})`).

That's just my first few discoveries.

wincent:
> That's just my first few discoveries.

Thanks for sharing your research!

[...]

PR #272, opened 2017-04-09
Add compatibility with Restify

santino:
This PR aims to address issues reported in #216.

## Goal
At the current stage `express-graphql` isn't really usable with restify and I believe that being able to claim full compatibility with it is a great improvement that will help a huge number of developers.

## What does this solve
1. Responses stringified twice
2. GraphiQL support

`express-graphql` is currently stringifying the reponse which is then also stringified by restify before being returned; this is causing the response to be strinfigied twice which would require double parsing on the requesting client application; this also breaks Relay implementation on the client side because Relay tries to parse the response once and expect to deal with an object with `data` and `error` properties.
Changes have been introduced to remove the stringification from `express-graphql` to then send the object response with `res.json()` rather than `res.send()` which is compatible with both `express` and `restify` so that they are explicitly asked to strinfigy the response.

`express-graphql` is responding to a GraphiQL request using `res.send()` which doesn't work with restify. Restify, infact, needs to be provided with a HTML string using `res.end()` in order to be processed as `text/html` type (which also needs to be set in the `Accept` headers when sending the request).
Using `res.end()` when sending `text/html` is still compatible with express as well as `connect`.

## Downside
By delegating the stringification of the response to the server (express/restify) we loose **the 'pretty' option** as we no longer control the stringification directly.
This is not really an issue considering that when using the GraphiQL interface the response is already rendered in a readable and useful state to the user; also many testing tools (like PostMan) have built-in options to pretty print JSON responses.

## Tests
The tests have been refactored to be more efficient and remove duplicated code.
All tests are now also executed against Restify as well.

## Additional notes
The README file has been updated to provide information of how to setup `express-graphql` with restify; also the 'pretty' option has been removed since is no longer supported.

### DIFF
--- a/src/__tests__/http-test.js
+++ b/src/__tests__/http-test.js
@@ -593,11 +593,6 @@ const basicConfig = [ urlString(), graphqlHTTP({ schema: TestSchema }) ];
 
       describe('allows for pre-parsed POST', () => {
         it('bodies', async function () {
-          // TODO: can we get this test working with restify?
-          if (name === 'restify') {
-            this.skip();
-          }
-
           // Note: this is not the only way to handle file uploads with GraphQL,
           // but it is terse and illustrative of using express-graphql and multer
           // together.
@@ -636,16 +631,20 @@ const basicConfig = [ urlString(), graphqlHTTP({ schema: TestSchema }) ];
 
           // Multer provides multipart form data parsing.
           const storage = multer.memoryStorage();
-          app.use(urlString(), multer({ storage }).single('file'));
-
-          // Providing the request as part of `rootValue` allows it to
-          // be accessible from within Schema resolve functions.
-          app.use(urlString(), graphqlHTTP(req => {
+          const multerHandler = multer({ storage }).single('file');
+          const config = [ urlString(), graphqlHTTP(req => {
             return {
               schema: TestMutationSchema,
               rootValue: { request: req }
             };
-          }));
+          }) ];
+
+          if (name === 'restify') {
+            app.post(config[0], multerHandler, config[1]);
+          } else {
+            app.use(config[0], multerHandler);
+            app.use(...config);
+          }
 
           const response = await request(app)
             .post(urlString())
--- a/src/__tests__/http-test.js
+++ b/src/__tests__/http-test.js
@@ -9,7 +9,7 @@
  */
 
 // 80+ char lines are useful in describe/it, so ignore in this file.
-// Also ignore no-unused-expressions rule as chai 'expect' sintax violates it.
+// Also ignore no-unused-expressions rule as chai 'expect' syntax violates it.
 /* eslint-disable max-len */
 /* eslint-disable no-unused-expressions */
 
--- a/src/__tests__/http-test.js
+++ b/src/__tests__/http-test.js
@@ -1255,8 +1255,7 @@ const basicConfig = [ urlString(), graphqlHTTP({ schema: TestSchema }) ];
       it('Do not execute a query if it do not pass the custom validation.', async() => {
         const config = [ urlString(), graphqlHTTP({
           schema: TestSchema,
-          validationRules: [ AlwaysInvalidRule ],
-          pretty: true,
+          validationRules: [ AlwaysInvalidRule ]
         }) ];
         if (name === 'restify') {
           app.get(...config);
--- a/src/__tests__/http-test.js
+++ b/src/__tests__/http-test.js
@@ -594,7 +594,7 @@ const basicConfig = [ urlString(), graphqlHTTP({ schema: TestSchema }) ];
       });
 
       describe('allows for pre-parsed POST', () => {
-        it('bodies', async function () {
+        it('bodies', async () => {
           // Note: this is not the only way to handle file uploads with GraphQL,
           // but it is terse and illustrative of using express-graphql and multer
           // together.
--- a/src/__tests__/http-test.js
+++ b/src/__tests__/http-test.js
@@ -15,14 +15,11 @@ import { expect } from 'chai';
 import { describe, it } from 'mocha';
 import sinon from 'sinon';
 import { stringify } from 'querystring';
-import url from 'url';
 import zlib from 'zlib';
 import multer from 'multer';
 import bodyParser from 'body-parser';
 import request from 'supertest-as-promised';
-import connect from 'connect';
-import express4 from 'express'; // modern
-import express3 from 'express3'; // old but commonly still used
+import express4 from 'express'; // current version 4
 import {
   GraphQLSchema,
   GraphQLObjectType,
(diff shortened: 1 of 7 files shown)


src/__tests__/http-test.js
```diff
@@ -9,20 +9,22 @@
  */
 
 // 80+ char lines are useful in describe/it, so ignore in this file.
+// Also ignore no-unused-expressions rule as chai 'expect' sintax violates it.
```
wincent: nit: sintax → syntax

santino: I'm afraid that might be too cryptic for me, do you want these 2 lines to be merged into one (will be looong); or would you flavour  a multi-line comment like so?
```
/*
Line 1
Line 2
*/
```

asiandrummer: @santino I think @wincent is referring to the typo - `sintax` should be `syntax` no?

santino: uh, oh, that would make sense!

wincent: Yeah, that's what I meant.

santino:
HI have fixed the typo, thanks for pointing that out.

@asiandrummer even if I believe no one should really have troubles with the missing `pretty` option I agree with you that this is indeed a breaking change which should be properly signalled with a bump to version 1 and a proper description in the changelog.

Unfortunately with the compatibility requirements I had in mind (express, restify and connect) there isn't any simple solution which is worth implementing; quite frankly the `pretty` option is just not worth the effort, the more Front End devs most likely are already using GraphiQL whilst the more Back End guys are probably using tools like PostMan to inspect GraphQL responses.

asiandrummer:
@santino to be honest I just don't know the consequences of this change well enough, especially removing `pretty` option. But we can work on this iteratively - we'll merge this and do a release with a proper version bump, and see if we need to bring it back.

BTW I love your documentation - thanks for the contribution!

src/index.js
```diff
@@ -309,10 +307,10 @@ function graphqlHTTP(options: Options): Middleware {
           operationName, result
         });
         response.setHeader('Content-Type', 'text/html; charset=utf-8');
-        sendResponse(response, payload);
+        response.end(payload);
       } else {
-        // Otherwise, present JSON directly.
-        const payload = JSON.stringify(result, null, pretty ? 2 : 0);
+        // Server will stringify our response object, we can return it directly
+        const payload = result || '';
```
leebyron: This is suspect. Why should we be returning empty string?

src/index.js
```diff
@@ -309,10 +307,10 @@ function graphqlHTTP(options: Options): Middleware {
           operationName, result
         });
         response.setHeader('Content-Type', 'text/html; charset=utf-8');
-        sendResponse(response, payload);
+        response.end(payload);
```
leebyron: why change to `response.end` here but leave `sendResponse` below?

leebyron:
This removes a ton of test coverage as well :(

leebyron:
Sorry, but I had to revert the landing of this PR. This should not be issued as a breaking change and we should not have so much altered test coverage. This is too wide of a change to the existing behavior.

I'm confident that with some minor additional changes, that this PR can be updated and not require these breaking changes.

src/index.js
santino: I use `response.end` because is a method available in all the supported servers (restify, express, connect) and it's just enough to return a response which we know can only be a string, since this is in response of a GraphiQL request.

The main purpose of `sendResponse` is to deal with different types of data and address them properly with the best server method available to handle them.
Initially I just kept the use of `response.send` in this function, but I know see that it can be removed which would also mean that I can restore the use of `sendResponse` also for the GraphiQL response, will update the PR.

src/index.js
santino: I agree that it's not ideal but I didn't want to change the existing logic with my PR as my purpose is just to bring compatibility with Restify.

The reason for the empty string is that at this level `result` can be `null` which is due to this  piece of code present 2 times, both at line 207 and 241
```
if (showGraphiQL) {
  return null;
}
```

If you don't want to alter the logic behind these 2 snippets then I have to deal with a possible `null` `result`.
The way your current code is dealing with that is to pass the `null` `result` to JSON.stringify like so `JSON.stringify(result, null, pretty ? 2 : 0)` which translates to `JSON.stringify(null, null, pretty ? 2 : 0)`; the result of this 'stringification' (which is an empty string) is then passed to `sendResponse`.
This means that I'm not introducing absolutely anything new by explicitly sending an empty string to `sendResponse` as this is exactly what was happening already, just it wasn't explicit and possibly not thought about/dealt with.

leebyron: Thanks for the analysis on this, I'll take a closer look

santino:
I'm afraid I'm not familiar with `isparta` and I'm having a few issues in getting it to work (I keep on getting `No coverage information was collected, exit without writing coverage information`).
I don't see why coverage should be altered by this PR but if you can point me to a meaningful report I will be happy to match the percentage to be the same as current master.

As about the breaking changes it's a bit disappointing to have the PR rolled back considering it's been sitting here for more than a month and it had just been merged...
As you can see from my comments I don't see an easy way to keep the `pretty` option if we use the proper `response.json` method available in express and restify and in all honesty I don't think this option is really worth the effort; I would love to hear some actual use cases to justify the necessity of such an option; after more than a month no-one so far has made any real comment on this.

leebyron:
> I don't see why coverage should be altered by this PR but if you can point me to a meaningful report I will be happy to match the percentage to be the same as current master.

I'm less concerned with the raw percentage coverage, and more concerned with the sequence of tests. Some tests were removed in the large refactoring of the tests in this PR.

> As about the breaking changes it's a bit disappointing to have the PR rolled back considering it's been sitting here for more than a month and it had just been merged

I definitely understand that. I'm sorry about the wait and the churn. We'll get all these issues addressed, I'm sure of it. Unfortunately github doesn't have great tools for readdressing a PR, so you may need to simply reopen a PR with the same stack of diffs (ideally rebased on latest master).

> As you can see from my comments I don't see an easy way to keep the pretty option if we use the proper response.json method available in express and restify and in all honesty I don't think this option is really worth the effort; I would love to hear some actual use cases to justify the necessity of such an option; after more than a month no-one so far has made any real comment on this.

It's used in some installations to ensure a human readable response on development servers. We should not break this behavior for people who rely on it. The easy way to keep it is to only use the proper `response.json` when `pretty` has not been provided. I think that should be pretty straight-forward and I can address quickly this in a separate PR which also addresses the missing response issue you pointed out earlier

santino:
I see your PR, thanks for issuing that.
Following your comments to my PR I'm now investigating solutions to keep the pretty option and testing those with my Restify server.

As about the tests I shall just say that the only removed tests are those regarding the removed pretty option.

I will be happy to make a new PR as soon as I'm happy with the changes I'm testing.

leebyron:
Hey @santino I was able to reapply your diff in e7a80014146683f138a4d7833c2f6af6b341d363 - I worked through the rebase on the recent master changes and in the process took a different approach with the tests which made for a much smaller change.

PR #273, opened 2017-04-11
Add favicon to prevent invalid GraphQL requests

wbyoung:
In my logs, I kept seeing `error: BadRequestError: Must provide query string.` and the request being made was for the favicon. This prevents that request from being made.

### DIFF
--- a/src/renderGraphiQL.js
+++ b/src/renderGraphiQL.js
@@ -62,6 +62,7 @@ add "&raw" to the end of the URL within a browser.
     }
   </style>
   <link href="//cdn.jsdelivr.net/graphiql/${GRAPHIQL_VERSION}/graphiql.css" rel="stylesheet" />
+  <link rel="shortcut icon" href="data:image/x-icon;," type="image/x-icon">
   <script src="//cdn.jsdelivr.net/fetch/0.9.0/fetch.min.js"></script>
   <script src="//cdn.jsdelivr.net/react/15.4.2/react.min.js"></script>
   <script src="//cdn.jsdelivr.net/react/15.4.2/react-dom.min.js"></script>


asiandrummer:
Thanks for the contribution! :D

leebyron:
Unfortunately this breaks existing favicon usage :(

leebyron:
Unfortunately I had to revert the merge of this PR since it was a breaking change.

A proper fix is to ensure that if you're exposing graphql at the root of your domain, that you're handling requests for other unrelated files before handling graphql requests. Other examples include `robots.txt`.

wbyoung:
@leebyron okay, that makes sense, but it may still be useful to somehow make it so this middleware will ignore requests that do make it though for these types of files. The [basic usage](https://github.com/graphql/express-graphql#simple-setup) explains how to set things up in a way that this type of error will be generated out of the box which isn't the best possible experience.

Any thoughts on where this could go?

Vanuan:
So the only way to fix it is to add some middleware which catches all requests except of root?

Vanuan:
So it's something like this:
```
app.use('/', handleNonRoot(), graphQLHTTP(req => ({
  schema,
  graphiql: true,
  formatError: error => formatError(error, req)
})));
```

But how do I know whether files exist or not? Do you mean to say I have to hard-code all the static assets of GraphiQL?

Vanuan:
For now this is good enough. But providing a default that causes errors is still not nice.
```
const handleNonRoot = function (req, res, next) {
  if(req.path === '/favicon.ico') {
    res.status(404)
   .send('Not found');
  } else {
    next()
  }
}

```

You see, the problem here is that when browser receives 400 when requesting favicon it would not calm down until it's 200 or 404.

Vanuan:
I believe a proper solution would be fixing GraphiQL that it would return 404 for all requests that are not GraphiQL static assets instead of serving 400 Bad Request.

Vanuan:
Hm, apparently GraphiQL doesn't have any static assets, does it? So this would work:

```
const handleNonRoot = function (req, res, next) {
  if(req.path !== '/') {
    res.status(404)
   .send('Not found');
  } else {
    next()
  }
}

```

But that would break other usages, for example when people deliberately want GraphiQL to serve all requests. Wait, why then in this case favicon.ico doesn't serve GraphiQL html?

PR #277, opened 2017-04-14
Add option for body size limit

ntietz:
This change is suggested due to an issue I ran into (client's data is larger than the body limit, in a way that I cannot separate into multiple requests) and that [others](http://stackoverflow.com/questions/41486401/error-request-entity-too-large-in-graphql-services-of-node) have also run into.

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -100,6 +100,12 @@ export type OptionsData = {
    * A boolean to optionally enable GraphiQL mode.
    */
   graphiql?: ?boolean,
+
+  /**
+   * An optional number to configure the maximum size of request body which is
+   * accepted by the server.
+   */
+  limit?: ?number,
 };
 
 /**
@@ -149,6 +155,7 @@ function graphqlHTTP(options: Options): Middleware {
     let graphiql;
     let formatErrorFn;
     let extensionsFn;
+    let limit;
     let showGraphiQL;
     let query;
     let documentAST;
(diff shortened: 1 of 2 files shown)


asiandrummer:
Interesting - may I ask what makes the request body so large? Typically I'd think 100kb is good enough for a request with queries/variables, but I'd like to understand this case a bit better.

ntietz:
In my case, my client's project has some very large documents that get stored into their database. Sometimes they can go over the 100kB limit when I do a mutation with one of these large documents. There are ways to design around it, but it seems like an unexpected and unnecessary limitation.

I definitely agree that, for all but the rarest cases, 100kB is well within limits. I don't see any harm in making it configurable, though, for those rare cases.

nodkz:
It will be very nice to allow tune `max_body_size`.

From my old school: suppose that Nginx or Apache keep its default maxBodySize limited by 2MB, so we will be restricted by uploading raw photos from mobile phones, can not upload big dumps to phpMyAdmin.

For now, when using graphql-express, I restricted to save 16Mb documents to MongoDB in graphql-compose-mongoose.  Or can not send batch update via graphql-compose-elasticsearch.

nodkz:
BTW `limit` is not a good name for option which restricted request size. Limits maybe ron request per second.

@leebyron while you're working today on `graphql-express` maybe you merge this PR manually with option renaming?

ntietz:
There was no activity on this to get it merged in, and there are now significant conflicts. Given that the original (non-conflicting) PR didn't get accepted, I don't want to invest the time to resolve the conflicts, so I'm just going to close this.

PR #288, opened 2017-04-25
Build .flow files and include package.json and fix types

jamiehodge:
Mimic the build scripts in https://github.com/graphql/graphql-js/blob/master/package.json#L30-L31.

- Include `.flow` and `package.json` files in build
- Fix incorrect type import (`$Response` from `express`)
- Replace `Request` type with `$Request` from `express`
- Ignore `/dist` when type checking

As a consumer of this module, it's annoying to not have the full request type. I can't see any good reason to maintain an internal `Request` type. Please correct me if I'm wrong.


wincent:
Thanks @jamiehodge. Like I said in https://github.com/graphql/graphql-relay-js/pull/155:

> I don't know if it will actually work, as I am not sure whether these `*.flow` files will get included in the distributed package, but it seems harmless enough to at least try it out. We'll find out if it works on the next release.

PR #304, opened 2017-05-16
Simplify Restify section in readme

wincent:
Then single example is probably enough. The preceding text makes it clear that you can choose to support `GET` or `POST` or both.

### DIFF
--- a/README.md
+++ b/README.md
@@ -49,7 +49,7 @@ app.post('/graphql', graphqlHTTP({
 
 app.get('/graphql', graphqlHTTP({
   schema: MyGraphQLSchema,
-  graphiql: true // set it to false to disable the GraphiQL interface
+  graphiql: true
 }));
 
 app.listen(4000);
--- a/README.md
+++ b/README.md
@@ -34,43 +34,7 @@ app.listen(4000);
 
 ## Setup with Restify
 
-In restify the `.use` method is reserved to common handlers so you have to manually register methods (GET, POST, etc) multiple times even for the same endpoint.
-
-If you just need GraphQL with POST requests, just mount `express-graphql` like so:
-
-```js
-const restify = require('restify');
-const graphqlHTTP = require('express-graphql');
-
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -49,7 +49,7 @@ app.post('/graphql', graphqlHTTP({
 
 app.get('/graphql', graphqlHTTP({
   schema: MyGraphQLSchema,
-  graphiql: true // set it to false to disable the GraphiQL interface
```
santino: the reason for this comment was to make it clear that setting `graphiql` to true is not mandatory just because this is a GET method.
Setting this to false will still allow to serve GraphQL requests through GET with query params.

wincent: I think the naming and preceding context makes it perfectly clear without the comment.

Without the comment, what would you expect `graphiql: false` to do? What would anybody expect it to do?

santino: I would expect people to understand what `graphiql` false does, but I also believe people will just copy/paste the code example provided without necessarily realising that option can be turned to false if they don't need GraphiQL.

wincent: I don't think we're worried about that. Look at all the other places in the README where we show `graphiql: true` without any comments. If we're concerned that this setting is non-obvious it should be renamed, and if we're concerned that mis-use of it is dangerous, then we've got the wrong API. But I don't think either of those are the case.

santino: I see your point and I don't seek to continue this conversation nor to push to re-introduce the comment.
The difference is that due to the nature of Restify you are required to setup GraphQL on multiple methods, the GET and the POST (if you need both, for example personally I'm ok with just the POST), this might cause confusion with the options that you pass in one or the other so I thought being a little bit more explicit could help and wouldn't harm.

wincent: I think the odds of confusion are extremely unlikely, and the downside even in the unlikely event of a misconfiguration is minimal, so I feel pretty happy about reducing the clutter in the example.

PR #315, opened 2017-05-27
move prettier to dev dependencies

charlieschwabacher:
I noticed installing the latest version downloaded prettier - i think it's safe to make this a dev dependency.


wincent:
Right you are. Excellent catch! Thanks @charlieschwabacher.

PR #318, opened 2017-05-31
Disable introspection query option

baer:
**What It Does**
This PR adds the option to apply an additional validation rule that disables query introspection. 

**Why**
In a production environment, allowing a 3rd party to run an inspection query is enough for them to figure out how to build any query they like. For the same reason that a company might want to turn off GraphiQL, they also might want to turn off this type of query. Ideally, queries are persisted at build time and GraphQL itself isn't accessible but there are a lot of people who expose GraphQL directly to consumers.

**How**
This PR adds to the existing GraphQL validation system by simply appending a new rule if the option is enabled.

### DIFF
--- a/README.md
+++ b/README.md
@@ -91,6 +91,8 @@ The `graphqlHTTP` function accepts the following options:
   * **`validationRules`**: Optional additional validation rules queries must
     satisfy in addition to those defined by the GraphQL spec.
 
+  * **`introspectionDisabled`**: If `true`, disable schema level introspection queries.
+
 In addition to an object defining each option, options can also be provided as
 a function (or async function) which returns this options object. This function
 is provided the arguments `(request, response, graphQLParams)` and is called
--- a/src/__tests__/disable-introspection-test.js
+++ b/src/__tests__/disable-introspection-test.js
@@ -1,0 +1,38 @@
+/**
+ *  Copyright (c) 2015, Facebook, Inc.
+ *  All rights reserved.
+ *
+ *  This source code is licensed under the BSD-style license found in the
+ *  LICENSE file in the root directory of this source tree. An additional grant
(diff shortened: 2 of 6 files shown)


albertstill:
This seems like a really sensible feature to add. I get security through obscurity is not ideal, but enabling this in prod seems like a quick and easy way to make life a bit harder for an attacker? 

I see Apollo has added [it](https://www.apollographql.com/docs/apollo-server/api/apollo-server/), also I see graphql-js has a [isIntrospectionType helper function](https://github.com/graphql/graphql-js/blob/c0cf320d3a569c83166ac93d36b2dfb81ebb5b75/src/index.js#L100).

albertstill:
Worth remembering GraphQL also gives these errors for wrong fields `"Cannot query field \"foox\" on type \"Query\". Did you mean \"foo\" or \"foos\"?"`

danielrearden:
@IvanGoncharov @acao Wondering if this wouldn't make more sense as a separate module. It's really just a validation rule, which is something that can be added using the existing API.

IvanGoncharov:
@danielrearden Don't think we need to create separate package for this.
I can add this validation rule into next version of `graphql-js` so it will be just a matter of adding it into the `validationRules`.

danielrearden:
@IvanGoncharov That would be awesome! I could open a PR over there if that would help.

IvanGoncharov:
> I could open a PR over there if that would help.

@danielrearden That would be great 👍
Please try to match the style and test of existing rules as close as possible.
Also, please create an issue against `graphql-js` so I can assign it to you.

IvanGoncharov:
Fixed by @danielrearden see this section for details:
https://github.com/graphql/express-graphql#disabling-introspection

PR #355, opened 2017-07-21
Allow formatError to be a promise

arunv:
Summary:
Often we need do so some datafetching (even from cache) to properly annotate the error.

Also useful if you want to log errors server-side to anything but the console output.

Test Plan:
First ran npm run test

Added the following route to the server:

When an error is thrown intentionally, we can see the "abe to process promises" logged. 
app.use(
  '/graphql',
  expressGraphQL(req => ({
    schema,
    graphiql: __DEV__,
    rootValue: { request: req },
    pretty: __DEV__,
    formatError: async function(error) {
          let ret = await testAsync();
          console.log(ret);
        return {
          working: 1,
          error: error,
        };
     }
  })),
);

async function testAsync() {
  return "able to process promises";
}

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -293,6 +293,18 @@ function graphqlHTTP(options: Options): Middleware {
         return { errors: [error] };
       })
       .then(result => {
+        // Format any encountered errors.
+        if (!result || !result.errors) {
+          return result;
+        }
+        return Promise.all(
+          result.errors.map(formatErrorFn || formatError).map(e => Promise.resolve(e))
+        ).then(errs => {
+          (result: any).errors = errs;
+          return result;
+        });
+      })
+      .then(function(result) {
         // If no data was included in the result, that indicates a runtime query
         // error, indicate as such with a generic status code.
         // Note: Information about the error itself will still be contained in
@@ -300,12 +312,6 @@ function graphqlHTTP(options: Options): Middleware {
         // http://facebook.github.io/graphql/#sec-Data
         if (result && result.data === null) {
           response.statusCode = 500;
-        }
-        // Format any encountered errors.
-        if (result && result.errors) {
-          (result: any).errors = result.errors.map(
-            formatErrorFn || formatError,
-          );
         }
 
         // If allowed to show GraphiQL, present it instead of JSON.


arunv:
Ah, I'm going to close this PR because the problem has been addressed. 

The proper way to solve this is:

app.use('/graphql', graphqlHTTP(async (request, response, graphQLParams) => ({
  schema: MyGraphQLSchema,
  rootValue: await someFunctionToGetRootValue(request)
  graphiql: true
})));

PatrLind:
I am not sure I understand the proper way of doing it as @arunv is suggesting.
The PR is about `formatError` to be able to take an async function, but the response seems not to to be about this at all?

arunv:
@PatrLind Since `formatError` is a property on an object that is returned by the function passed to graphqlHTTP, you can make that function be async. 

So, in my example what is happening is:

```
app.use('/graphql', expressGraphQL(req => ({ 
  schema, 
  graphiql: true, 
  formatError: formatErrorFn 
  /* .. etc etc ...*/
})))
```

This is the same as:

```
app.use('/graphql', expressGraphQL(req => {
return { 
  schema, 
  graphiql: true, 
  formatError: formatErrorFn 
  /* .. etc etc ...*/
};
})))
```
This is the same as:

```
app.use('/graphql', expressGraphQL(function(req) {
return { 
  schema, 
  graphiql: true, 
  formatError: formatErrorFn 
  /* .. etc etc ...*/
};
})))
```

Now all you have to do is convert this function into an async function

```
app.use('/graphql', expressGraphQL(async function(req) {
return { 
  schema, 
  graphiql: true, 
  formatError: formatErrorFn 
  /* .. etc etc ...*/
};
})))
```
Now that it's an async function, you can await another async function inside of it

```
app.use('/graphql', expressGraphQL(async function(req) {
return { 
  schema, 
  graphiql: true, 
  formatError: await asyncFormatErrorFn(req),
  /* .. etc etc ...*/
};
})))
```

PatrLind:
@arunv Thanks for the suggestion, however, I am not sure I understand how this will solve the problem.
If I await like you suggest, that awaits the function pointer, not the processed error. The error processing still needs to be done synchronously. Right?
It seems to be that your original PR would have fixed this issue.

Issue #388, opened 2017-08-29
Reduce verbosity of Travis → NPM deploy log output


wincent:
As noted in #385, it is now too verbose; see: https://travis-ci.org/graphql/express-graphql/jobs/269661199

PR #389, opened 2017-08-29
Silence npm install output in Travis → NPM deploy

wincent:
Without this, the `npm install` output causes the logs to be truncated:

https://travis-ci.org/graphql/express-graphql/jobs/269661199

Note that I am using an old-school POSIX-compatible redirect of stderr and stdout (as opposed to the simpler `npm install &> /dev/null` because I don't know whether this script is going to be running as Bash or `/bin/sh` (which itself could be Bash, or something else).

Closes: https://github.com/graphql/express-graphql/issues/388

### DIFF
--- a/resources/prepublish.sh
+++ b/resources/prepublish.sh
@@ -17,5 +17,5 @@ if [ "$CI" != true ]; then
 fi
 
 # Build before Travis CI publishes to NPM
-npm install
+npm install > /dev/null 2>&1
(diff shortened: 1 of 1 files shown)


wincent:
Can't test this without actually cutting a release, but did do a `npm install > /dev/null 2>&1` locally to confirm that I got the syntax right.

PR #391, opened 2017-08-31
Allow to replace default execute function

IvanGoncharov:
This PR allows providing your own `execute` function instead of the default one from the `graphql-js`.
It’s a small change that adds a lot of flexibility since you can wrap standard `execute` and it opens new possibilities for middlewares. 

Personally, I need it to proxy calls to 3rd-party APIs.
It’s an alternative to #253 but introduces less code and more flexible mechanism.
It’s very important for us as it will be the last change required to switch [graphql-faker](https://github.com/APIs-guru/graphql-faker) from custom forks of `graphql-js` and `graphql-express`.
@wincent @leebyron Can you please review it?

### DIFF
--- a/README.md
+++ b/README.md
@@ -90,6 +90,9 @@ The `graphqlHTTP` function accepts the following options:
   * **`customValidateFn`**: An optional function which will be used to validate
     instead of default `validate` from `graphql-js`.
 
+  * **`customExecuteFn`**: An optional function which will be used to execute
+    instead of default `execute` from `graphql-js`.
+
   * **`customFormatErrorFn`**:  An optional function which will be used to format any
     errors produced by fulfilling a GraphQL operation. If no function is
     provided, GraphQL's default spec-compliant [`formatError`][] function will be used.
--- a/src/__tests__/http-test.js
+++ b/src/__tests__/http-test.js
@@ -29,6 +29,7 @@ import {
   GraphQLError,
   BREAK,
   validate,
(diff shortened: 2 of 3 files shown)


matthewerwin:
@IvanGoncharov why has the signature been changed to a single args object? Public API that you cited on my PR http://graphql.org/graphql-js/execution/#execute doesn't take a single object argument so this seems like this code isn't working when the override execute isn't passed in.

IvanGoncharov:
@matthewerwin `execute` was extended to also support object as argument see here:
https://github.com/graphql/graphql-js/pull/867
But no one updated the official docs 😞 

Also, my PR to expose `ExecutionArgs` type was merged: https://github.com/graphql/graphql-js/pull/988
So it's easy to do type checks in Flow and TS.

I think receiving/returning objects is better than 7 separate values.

matthewerwin:
@IvanGoncharov right on. Maybe make the update to the docs on that part of this PR. Thanks for the comprehensive info

matthewerwin:
@IvanGoncharov I think before this PR gets merged it would be great to add two things:

1. The documentation updates.
2. change "response.statusCode = 400;" to "response.statusCode = contextError.status||400;"

First one to ensure everyone understands the execute capability, the second so you can send back meaningful codes (e.g. 401 for unauthorized).

IvanGoncharov:
@leebyron Can you please review this PR?

PR #395, opened 2017-09-28
Readable stack trace ftw instead of a difficult to read one liner

soneymathew:
A minor update to help users navigate the errors
image

previously it was a one liner

image


IvanGoncharov:
@soneymathew Great idea 👍 
Merged 🎉 Sorry for the delay.

PR #402, opened 2017-11-08
Fix a typo

OriginalEXE:


IvanGoncharov:
@OriginalEXE Merged 🎉 Thanks for the fix and sorry for delay.

PR #410, opened 2018-01-16
remove inline js

graingert:


IvanGoncharov:
@graingert I don't think moving JS in separate file such big improvement to justify added complexity and  new dependency on`escape-html`.

P.S. Personally, I think such code should live inside `graphiql/renderHTMLPage` so it doesn't duplicated in this and other graphql servers (e.g. [apollo-server](https://github.com/apollographql/apollo-server/blob/master/packages/apollo-server-module-graphiql/src/renderGraphiQL.ts)).

IvanGoncharov:
@graingert Thanks for PR but I'm closing it because it became outdated after #524

PR #416, opened 2018-02-10
Fix graphQl promises in IE11

311289189:


IvanGoncharov:
@311289189 Thanks for PR 👍
I copied the solution from GraphiQL example:
https://github.com/graphql/graphiql/blob/master/example/index.html#L30
And committed it as 5a082b2a1d02aadb1aa5a284e61f03aaf7fc37de

IvanGoncharov:
@311289189 IE 10/11 should work in `v0.6.12` 📦  
Would be great if you or someone else will test it since I don't have access to Windows machine.

PR #426, opened 2018-04-13
Custom error formatting does not work when parameters are malformed

heneryville:
When any of the core GraphQL parameters are malformed, custom error formatting stops working, and instead errors are returned in the default GraphQL flavor.

This happens for all of the following errors:
- POST body sent invalid JSON.
- Unsupported charset
- Invalid body
- Unsupported content-encoding
- Variables are invalid JSON
- Any other errors that occur when parsing body, e.g. uncompressing errors

When parameters are malformed the code to set`formatErrFn` is [never reached](https://github.com/graphql/express-graphql/blob/master/src/index.js#L181), so errors fallback to the default GraphQL error function.

This change will instead invoke the options parsing function when parameters fail to parse with the specific intention of extracting the `formatError` function for use in reporting the error. A dummy parameters (with all values set to null) is passed to the options function, since it seems likely that existing clients rely on the object itself existing.

### DIFF
--- a/src/__tests__/http-test.js
+++ b/src/__tests__/http-test.js
@@ -1433,6 +1433,37 @@ describe('test harness', () => {
         });
       });
 
+      it('allows for custom error formatting of poorly formed requests', async () => {
+        const app = server();
+
+        get(
+          app,
+          urlString(),
+          graphqlHTTP({
+            schema: TestSchema,
+            formatError(error) {
+              return { message: 'Custom error format: ' + error.message };
+            },
+          }),
+        );
+
+        const response = await request(app).get(
+          urlString({
+            variables: 'who:You',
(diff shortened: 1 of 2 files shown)


heneryville:
Signed

IvanGoncharov:
@heneryville Thanks for RP 👍 
I like your PR but I want to handle a few additional cases e.g. `pretty` option, validate that `options` return object. So I want to reuse existing code and that's why I extracted common part into `resolveOptions` function.

I committed my changes on top of your PR so could you please review them?

IvanGoncharov:
@heneryville Can you please review it?

IvanGoncharov:
> Yes! I was out all last week, so missed this, but your changes look great. I'm glad they got merged.

No worries. 
Thank you for your work on this PR 👍

Issue #425, opened 2018-04-07
use with ES6 modules triggers assertion in graphql


stevvvn:
The following does not work, though it seems to me it should (node@9.8.0):
```js
import graphqlHTTP from 'express-graphql'; // @0.6.12
import { buildSchema } from 'graphql'; // @1.13.2
import express from 'express';

const app = express();
app.use('/graphql', graphqlHTTP({
  'schema': buildSchema('type Query { hello: String }'),
  'rootValue': { 'hello': () => 'Hello, world!' }
}));
app.listen(4000);
```

```
$ curl -s -X POST -H "Content-type: application/json" http://localhost:4000/graphql --data '{ "query": "{ hello }" }'|jq
{
  "errors": [
    {
      "message": "Cannot use GraphQLSchema \"[object Object]\" from another module or realm.\n\nEnsure that there is only one instance of \"graphql\" in the node_modules\ndirectory. If different versions of \"graphql\" are the dependencies of other\nrelied on modules, use \"resolutions\" to ensure only one version is installed.\n\nhttps://yarnpkg.com/en/docs/selective-version-resolutions\n\nDuplicate \"graphql\" modules cannot be used at the same time since different\nversions may have different capabilities and behavior. The data from one\nversion used in the function from another could produce confusing and\nspurious results."
    }
  ]
}
```

Creating a wrapper to `require` buildSchema instead:

**req.js**
```js
module.exports = require('graphql').buildSchema;
```
And changing the initial example:
```js
//import { buildSchema } from 'graphql';
import buildSchema from './req';
```

works as expected:
```
$ curl -s -X POST -H "Content-type: application/json" http://localhost:4000/graphql --data '{ "query": "{ hello }" }'|jq
{
  "data": {
    "hello": "Hello, world!"
  }
}
```

scinos:
I've been investigating this and I don't have a solution, but I'll share what I found.

The core of the problem is that there are two different ways to get a `GraphQLSchema`:

* `import { buildSchema } from 'graphql'` ends up using `node_modules/graphql/type/schema.mjs` (an ES6 module).
* `import graphqlHTTP from 'express-graphql'` does `require('graphql')`, which ends up loading `node_modules/graphql/type/schema.js` (a CJS module).

[...]

PR #433, opened 2018-05-02
Support native ESM via .mjs.

jaydenseric:
[Node.js natively supports ESM via `.mjs`](https://nodejs.org/api/esm.html#esm_ecmascript_modules) (currently behind the `--experimental-modules` flag, soon to be unflagged).

At the moment, consumers using native ESM, `graphql` and `express-graphql` are in a conundrum (see https://github.com/graphql/express-graphql/issues/425). When they use `graphql` the ESM `.mjs` modules run. When `express-graphql` internally uses `graphql`, the CJS `.js` files run. `graphql` thinks there are multiple versions being used and throws a tantrum. This issue will be mitigated by supporting native ESM across both packages.

- Swapped deprecated `babel-preset-es2015` for `babel-preset-env`.
- Added `package.json` `engines` field.
- Configured `babel-preset-env` to target the Node.js version defined in `package.json` `engines` field.
- Kept Babel at v6, but moved config into a v7 ready `.babelrc.js` file for dynamic config via env.
- Tidied the `package.json` scripts, added an ESM build step that generates `.mjs` dist files. Follows the lead of https://github.com/graphql/graphql-js/pull/1244, although I do things quite different for my own projects with Babel v7.
- Updated `package.json` `main` field for ESM/CJS cross compatibility.
- Added a `package.json` `module` field for tree-shaking bundlers.

I didn't upgrade Babel to v7 to limit the scope of this PR. I do think it's a good idea though.

### DIFF
--- a/.babelrc.js
+++ b/.babelrc.js
@@ -1,0 +1,15 @@
+module.exports = {
+  presets: [
+    [
+      'env',
+      {
+        modules: process.env.ESM ? false : 'commonjs',
+        targets: {
+          node: require('./package.json').engines.node.substring(2), // Strip `>=`
+        },
+      },
+    ],
+  ],
+  plugins: ['transform-class-properties', 'transform-flow-strip-types'],
+  ignore: ['__tests__'],
+};
--- a/package.json
+++ b/package.json
@@ -19,7 +19,11 @@
     "middleware",
     "api"
   ],
-  "main": "dist/index.js",
+  "engines": {
+    "node": ">=4"
+  },
+  "main": "dist/index",
+  "module": "dist/index.mjs",
   "directories": {
     "lib": "./dist"
   },
@@ -34,26 +38,25 @@
   },
   "babel": {
     "presets": [
-      "es2015"
-    ],
-    "plugins": [
-      "transform-class-properties",
-      "transform-flow-strip-types"
+      "./.babelrc.js"
     ]
   },
   "scripts": {
-    "prepublish": ". ./resources/prepublish.sh",
     "test": "npm run lint && npm run check && npm run testonly",
     "testonly": "mocha $npm_package_options_mocha",
     "lint": "eslint src",
     "prettier": "prettier --write 'src/**/*.js'",
     "check": "flow check",
-    "build": "rm -rf dist/* && babel src --ignore __tests__ --out-dir dist && npm run build:flow",
+    "build": "npm run build:clean && npm run build:esm && npm run build:cjs && npm run build:flow",
+    "build:clean": "rm -rf dist/*",
+    "build:esm": "ESM=1 babel src -d dist/esm && for file in $(find dist/esm -name '*.js'); do mv \"$file\" `echo \"$file\" | sed 's/dist\\/esm/dist/g; s/.js$/.mjs/g'`; done && rm -rf dist/esm",
+    "build:cjs": "babel src -d dist",
     "build:flow": "find ./src -name '*.js' -not -path '*/__tests__*' | while read filepath; do cp $filepath `echo $filepath | sed 's/\\/src\\//\\/dist\\//g'`.flow; done",
     "watch": "node resources/watch.js",
     "cover": "babel-node node_modules/.bin/isparta cover --root src --report html node_modules/.bin/_mocha -- $npm_package_options_mocha",
     "cover:lcov": "babel-node node_modules/.bin/isparta cover --root src --report lcovonly node_modules/.bin/_mocha -- $npm_package_options_mocha",
     "preversion": "npm test",
+    "prepublish": ". ./resources/prepublish.sh",
     "start": "babel-node examples/index.js"
   },
   "dependencies": {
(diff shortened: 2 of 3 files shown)


jaydenseric:
Don't merge yet, I'm noticing that `dist/index.mjs` has a few erroneous `module.exports` in it for some reason.

IvanGoncharov:
Looks great 🎉 

> Don't merge yet, I'm noticing that dist/index.mjs has a few erroneous module.exports in it for some reason.

Please ping me when you solve this issue or if you need help with it.
I was planning on doing `0.7.0` at the end of this week and would be great to also include this feature.

> Tidied the package.json scripts, added an ESM build step that generates .mjs dist files. Follows the lead of graphql/graphql-js#1244, although I do things quite different for my own projects with Babel v7.

If you have time you can help to migrate `graphql-js`: https://github.com/graphql/graphql-js/issues/1266
It would be ideal to first update it and than all other `graphql/*` projects.

> Added a package.json module field for tree-shaking bundlers.

Maybe it also make sense to add `"sideEffects": false` simular to https://github.com/graphql/graphql-js/pull/1312

> Added package.json engines field.

No need to support `4.0` since it's `End-of-life` so you can set it to `6.0`
https://github.com/nodejs/Release

jaydenseric:
As a breaking change, I updated the exports to be only named for ESM/CJS cross compatibility and consistency.

For a native ESM environment:

```js
import { graphqlHTTP, getGraphQLParams } from 'express-graphql'
```

For a legacy CJS environment:

```diff
- const graphqlHTTP = require('express-graphql')
- const getGraphQLParams = graphqlHTTP.getGraphQLParams
+ const { graphqlHTTP, getGraphQLParams } = require('express-graphql')
```

The old way the additional `getGraphQLParams` export was added onto the `graphqlHTTP` function was pretty weird anyway, and now the `import` and `require` look consistent.

There is no concept of mixed default/named exports in CJS, only a babel convention of a `.default` export. If we had mixed, this is what the API would look like:

```js
// ESM:
import graphqlHTTP, { getGraphQLParams } from 'express-graphql'

// CJS:
const { default: graphqlHTTP, getGraphQLParams } = require('express-graphql')
```

.travis.yml
```diff
@@ -2,10 +2,8 @@ language: node_js
 
 # https://github.com/nodejs/LTS
 node_js:
-  - "8"
```
IvanGoncharov: `8` is still in `Active LTS` and we should definitely test it + we need to test `10`

.babelrc.js
```diff
@@ -0,0 +1,14 @@
+module.exports = {
+  presets: [
+    [
+      'env',
+      {
+        modules: process.env.ESM ? false : 'commonjs',
+        targets: {
+          node: require('./package.json').engines.node.substring(2), // Strip `>=`
```
IvanGoncharov: Until https://github.com/babel/babel/issues/7277 is implemented I think it's better to use explicit string.

jaydenseric: Why? Then you would not have a single source of truth.

.travis.yml
jaydenseric: v10 is tested as it is the current stable release used via `'node'`.

IvanGoncharov:
> As a breaking change, I updated the exports to be only named for ESM/CJS cross compatibility and consistency.

I would prefer not make such breaking changes especially since the absolute majority of users don't use `getGraphQLParams` function.
Can we make it backward compatible using [babel-plugin-add-module-exports](https://github.com/59naga/babel-plugin-add-module-exports)?

.babelrc.js
IvanGoncharov: Because this construct assumes `>= <version>` and is very brittle.
I personally think `engines` should contain only node version that we run tests against. See my PR for `graphql-js`: https://github.com/graphql/graphql-js/pull/1338/files#diff-b9cfc7f2cdf78a7f4b91a753d10865a2R21

.travis.yml
IvanGoncharov: I prefer to use explicit version, especially for consistency with `graphql-js` and other `graphql/*` repos.

jaydenseric:
@IvanGoncharov

> Can we make it backward compatible using babel-plugin-add-module-exports?

For starters, [that plugin does not, and likely will not support babel v7](https://github.com/59naga/babel-plugin-add-module-exports/issues/51).

The best way forward is to tweak the API to properly support both environments. As a best practice, no npm package should have mixed named/default exports. Only default, or only named.

The new API would be consistent with `graphql` and `graphql-relay`, which only have named exports.

This issue is a dealbreaker for me and others I know. Because Apollo does not support native ESM either I've begun work on a new GraphQL server package.

jdalton:
>  (currently behind the --experimental-modules flag, soon to be unflagged).

Hi 👋 ! Node core and Node Module WG member here. There is no set date for when `--experimental-modules` will be unflagged. There's also no guarantee that what has shipped now is what will be in the future. For example, there are several discussions in the Node Module WG on how to approach implementation. One such discussion is to ship a maximally minimal implementation that would require package-maps for resolving bare specifiers (no longer using the Node module resolution algorithm). 

Adopting `.mjs` with the idea that it's current support is solid (for production use) at this stage is bit premature and may cause headache for users. Not only #425 but also things like for example Webpack will lock down bundled `.mjs` bits so that mocks, stubs, spys no longer work on things like transpiled namespace objects. Or if/when changes do happen to Node's `--experimental-modules` your use ESM story will be various shades of broken.

The `--experimental-modules` flag really does mean experimental. It's something you should pause and consider at least before going all in that route for your ESM support strategy today.

jaydenseric:
Things were moving too slow so I published [graphql-api-koa](https://github.com/jaydenseric/graphql-api-koa), and have been enjoying writing resolvers in native ESM for a month now without issue.

FredrikZeiner:
We're about a month away from Node 12 going into LTS. What happened to this feature?

PR #441, opened 2018-05-31
Add context to the arguments of the extensions function

corey-clark:
This information can be very useful when adding the extensions property to the response.

### DIFF
--- a/README.md
+++ b/README.md
@@ -86,7 +86,7 @@ The `graphqlHTTP` function accepts the following options:
     `"extensions"` field in the resulting JSON. This is often a useful place to
     add development time metadata such as the runtime of a query or the amount
     of resources consumed. This may be an async function. The function is
-    given one object as an argument: `{ document, variables, operationName, result }`.
+    given one object as an argument: `{ document, variables, operationName, result, context }`.
 
   * **`validationRules`**: Optional additional validation rules queries must
     satisfy in addition to those defined by the GraphQL spec.
@@ -201,7 +201,7 @@ must return a JSON-serializable Object.
 When called, this is provided an argument which you can use to get information
 about the GraphQL request:
 
-`{ document, variables, operationName, result }`
+`{ document, variables, operationName, result, context }`
 
 This example illustrates adding the amount of time consumed by running the
 provided query, which could perhaps be used by your development tools.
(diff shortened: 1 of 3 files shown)


IvanGoncharov:
@corey-clark Great PR 👍 
Can you please add tests?

corey-clark:
@IvanGoncharov unit test added 🙂

README.md
```diff
@@ -218,7 +218,7 @@ app.use('/graphql', graphqlHTTP(request => {
   return {
     schema: MyGraphQLSchema,
     graphiql: true,
-    extensions({ document, variables, operationName, result }) {
+    extensions({ document, variables, operationName, result, context }) {
```
IvanGoncharov: Would be great if you put `runTime` in context and make `extension` into a function defined on a top level. It will give realistic usage example for `context` argument

src/__tests__/http-test.js
```diff
@@ -1932,6 +1932,37 @@ describe('test harness', () => {
         );
       });
 
+      it('should have access to the GraphQL context property', async () => {
```
IvanGoncharov: Looking at above test I think it's better to modify it similar to my suggestion about the example in README, instead of adding new test specifically for context.

src/index.js
```diff
@@ -131,6 +131,11 @@ export type RequestInfo = {
    * The result of executing the operation.
    */
   result: ?mixed,
+
+  /**
+   * A value to pass as the context to the graphql() function.
```
IvanGoncharov: I think the reference to `graphql` is confusing maybe just `The value passed as the context to this middleware` or something similar. What do you think?

corey-clark: Sure, that sounds fine to me too.

corey-clark:
@IvanGoncharov made changes based on your suggestions, hope this is what you had in mind.

IvanGoncharov:
@corey-clark Thanks 👍 Merged 🎉 
I was planing to do next release after `graphql-js@14.0.0` released.

fernandocalsa:
Hi, thank you for this code, is exactly what I was looking for, but do you know when is it going to be released? It is in the Readme and I spent some time digging out why I didn't receive the context in the extension function.

Thank you!

donacross:
Thank you for this great feature. I had the exact same confusion as @fernandocalsa [regarding the doc](https://github.com/graphql/express-graphql#providing-extensions).

I just wanted to point out that you may want to update the [DefinitelyTyped repo](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/434dfb036dffba2b6787c1370ab93f90299d1b55/types/express-graphql/index.d.ts#L83) alongside this PR.

Thanks a lot for your great work

PR #443, opened 2018-06-13
Allow handler error with an express middleware

adriexnet:
### What does this PR do?
Allow using an express middleware as handler error

### How should this be tested?
• Configure graphqlHTTP with handleErrorWithExpress
```
app.use('/graph', graphqlHTTP(async (req) => {
  return {
    graphiql: true,
    schema,
    handleErrorWithExpress: true,
    formatError(error) {
      logger.error(error.message, {
        path: error.path,
        locations: error.locations,
        stack: error.stack ? error.stack.split('\n') : [],
        trace: error.trace,
      });
      return {
        message: error.message,
        status: 400,
      };
    }
  };
}));
```
• Add an error middleware
```
app.use((error, req, res, next) => {
  if (res.headersSent) {
    return next(error);
  }

  return res.status(error.status || 500).send({
    error: error.message
  });
});
```

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -32,7 +32,7 @@ import type {
   GraphQLSchema,
   GraphQLFieldResolver,
 } from 'graphql';
-import type { $Request, $Response } from 'express';
+import type { $Request, $Response, $NextFunction } from 'express';
 
 /**
  * Used to configure the graphqlHTTP middleware by providing a schema
@@ -69,6 +69,11 @@ export type OptionsData = {
    * A boolean to configure whether the output should be pretty-printed.
    */
   pretty?: ?boolean,
+
+  /**
+   * A boolean to configure whether an error should be handler by an express middleware.
+   */
+  handleErrorWithExpress?: ?boolean,
 
   /**
    * An optional function which will be used to format any errors produced by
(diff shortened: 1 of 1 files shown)


ValfarDeveloper:
Oh man this is so useful! It's a shame that yet is not implemented.

src/index.js
```diff
@@ -336,6 +350,16 @@ function graphqlHTTP(options: Options): Middleware {
...
         }
 
+        // If handler error with express is allowed, return the first error in next function.
+        if (
+          handleErrorWithExpress &&
+          result &&
+          result.errors &&
+          result.errors.length
+        ) {
+          return nextFunction(result.errors[0]);
```
jsonmaur: I'd say we should return all the errors in the array, not just the first one.

aagtt:
for god sake, accept this!

driescroons:
@aagtt @jsonmaur 

I wouldn't add the ability to add error middleware to the package itself. I've handled it like so:
- I use express-async-errors (https://www.npmjs.com/package/express-async-errors)
- I throw the error in my customFormatErrorFn
- My error middleware catches the error and is able to handle it according to the type

Here a demo:
https://streamable.com/hvtjg3

PR #451, opened 2018-09-17
WIP: allow the user to disable HTTP 500 if the response contains only errors

julkwiec:
By default, the `express-graphql` library changes the HTTP response code to 500 if the `response.data` property is falsy.
As a back-end developer, I want to be able to decide whether to respond with HTTP 200 or HTTP 500, even if the response contains no `data`, just `errors`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -77,6 +77,8 @@ The `graphqlHTTP` function accepts the following options:
 
   * **`pretty`**: If `true`, any JSON response will be pretty-printed.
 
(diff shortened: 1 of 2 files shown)


nvieirarafael-sensedia:
This option would be great! 
I recently update my project to use latest version of the lib and these 500 status code on queries without data are giving me integration problems with clients that already use api.

julkwiec:
Fixed merge conflicts.

Also, #471 seems like a better solution ;)

Issue #427, opened 2018-04-13
Queries returning only errors are forced to be a 500


derek-miller:
On commit [b3ccce9](https://github.com/graphql/express-graphql/commit/b3ccce9e0ab0acad122522cb7860d54731f5e8d5) a query response containing only the `errors` field is forced to be a 500, however these errors could easily be coercion errors due to incorrect input types, thus a bad request and not a server error. I am wondering if there is a strong opinion to keep it this way or if we could improve this logic and potentially make it configurable? A lot of client code that makes graphQL requests often retry on 500 errors and in the case of coercion errors will never succeed.

andfk:
+1 having an issue doing proper error handling in client-side as it always returns status 500

masiamj:
@andfk Can you elaborate on your solution? What version are you using now? How did it change?

Best!

andfk:
hey @masiamj yeah. I think i may have to edit my comment as is wrong. I initially thought the error was coming from Apollo and it'll be solved by upgrading `apollo-link-http` anyway the issue still remains. What i ended doing as a temporal solution (i hope) was to remove the '!' or 'new GraphqlNonNull()' from the responses so the 500 is not returned if the response is empty when it has errors. For example a user error we throw manually not an unhandled expection or so.

Hope that helps and lets see how this issue goes. I personally like much more the previous approach.

julkwiec:
I'm facing the same issue. I tracked it down to [this line in the express-graphql library](https://github.com/graphql/express-graphql/blob/master/src/index.js#L332);

Basically, if the GraphQL response data (`result.data`) is falsy, and the HTTP response code is set to 200, it's being set to 500 instead.

I'd love to have a possibility to change this behavior.

egustavson:
+1

berstend:
Hardcoded 5xx error made me a little sad, as this might confuse certain GraphQL clients.

This [PR](https://github.com/graphql/express-graphql/pull/451) doesn't seem to be exhaustive or about to be merged and I also didn't feel like maintaining a fork.

I therefore resorted to the next best thing, hijacking the express send handler. 🐴 

```typescript
import { NextFunction, Request, Response } from "express";
import * as hijackResponse from "hijackresponse";

// Extend Express Response with hijack specific function
interface IHijackedResponse extends Response {
  unhijack: () => void;
}

/**
 * Stupid problems sometimes require stupid solutions.
[...]

PR #471, opened 2019-02-05
Alternative fix for #427

rogierschouten:
Issue #427 already has a PR #451 but this solution is a bit more flexible. 

This fixes notably the issue that when the client forgets a mandatory field in a GraphQL mutation input,
the underlying graphql-js package does not throw; rather it returns a normal response with errors. This should result in status code 400 (bad request) because it is a client problem, but express-graphql makes it a 500. 

This PR allows to specify separately what to do with normal responses that
- contain no data and no errors (typically 500)
- contain no data and errors (typically 400)
- contain both data and errors (typically 200)

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -108,6 +108,21 @@ export type OptionsData = {
    * value or method on the source value with the field's name).
    */
   fieldResolver?: ?GraphQLFieldResolver<any, any>,
+
+  /**
+   * Status code for results that have neither errors nor data; default 500
+   */
+  noErrorNoDataStatusCode?: number;
+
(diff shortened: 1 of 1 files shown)


jsonmaur:
@rogierschouten I was about to open a similar PR for this. It really needs to be fixed... issues with the data coming from clients should never return a 500 status code.

@IvanGoncharov Can we get this merged?

jsonmaur:
After looking at other PR's, [this idea](https://github.com/graphql/express-graphql/pull/443) of using Express middleware might be a better approach.

Issue #474, opened 2019-02-26
Performance improvement caching validation


jamesmoriarty:
I'm currently using GraphQL to transform large documents and I've noticed a significant amount of time in validate. https://github.com/graphql/express-graphql/blob/master/src/index.js#L254

```
validate(schema, documentAST, validationRules);
```

The time spent on the first request:

```
parse: 21.042ms
hash: 3.727ms
validate: 965.500ms <
execute: 331.606ms
```

If I cache the result of validate. The time spent on the second request:

```
parse: 49.794ms
hash: 5.105ms
validate: 5.203ms <
execute: 336.634ms
```

As far as I can tell the schema is static, the query is the same with different arguments, and the validation rules are the same.

The change looks something like:

```
hash = hashFn(schema, documentAST, validationRules)

if (!cache[hash]) {
   cache[hash] = validate(schema, documentAST, validationRules);
}

const validationErrors = cache[hash];
```

As it's currently stateless, I'm trying to understand the trade-offs caching the result. I would appreciate any input from the contributors / maintainers. Basically, is this a terrible idea?

IvanGoncharov:
@jamesmoriarty GraphQL validate functionality was designed with a pretty similar case in mind:
https://facebook.github.io/graphql/June2018/#sec-Validation
> Typically validation is performed in the context of a request immediately before execution, however a GraphQL service may execute a request without explicitly validating it if that exact same request is known to have been validated before. For example: the request may be validated during development, provided it does not later change, or a service may validate a request once and memoize the result to avoid validating the same request again in the future. Any client‐side or development‐time tool should report validation errors and not allow the formulation or execution of requests known to be invalid at that given point in time.

Moreover you can optimise it a little bit more by hashing query string before parsing it to AST.
AFAIK Facebook engineer went one step forward and hash queries during development  so client sents only hash. Here is a talk with more details: https://youtu.be/G_zipR8Y8Ks

jamesmoriarty:
Thanks @IvanGoncharov. I really appreciate the response. 

Is this a feature the maintainers would be interested in merging? If so is there any guidence on how the maintainers would ideally like it implemented? 

```
app.use('/graphql', graphqlHTTP({
  schema: MyGraphQLSchema,
  graphiql: true,
  validationCache: false <default> | <cache object e.g. {} or LRU>
}));
```

IvanGoncharov:
Fixed in master.
@jamesmoriarty Thanks for PR 👍

[...]

PR #479, opened 2019-03-20
refine custom operations. re: #474

jamesmoriarty:
Regarding: #474

### DIFF
--- a/README.md
+++ b/README.md
@@ -87,14 +87,14 @@ The `graphqlHTTP` function accepts the following options:
   * **`validationRules`**: Optional additional validation rules queries must
     satisfy in addition to those defined by the GraphQL spec.
 
-  * **`customValidateFn`**: An optional function which will be used to validate 
+  * **`customValidateFn`**: An optional function which will be used to validate
     instead of default `validate` from `graphql-js`.
 
   * **`customFormatErrorFn`**:  An optional function which will be used to format any
     errors produced by fulfilling a GraphQL operation. If no function is
     provided, GraphQL's default spec-compliant [`formatError`][] function will be used.
 
-  * **`formatError`**: is deprecated and replaced by `customFormatErrorFn`. It will be 
+  * **`formatError`**: is deprecated and replaced by `customFormatErrorFn`. It will be
     removed in version 1.0.0.
 
 In addition to an object defining each option, options can also be provided as
--- a/README.md
+++ b/README.md
@@ -77,10 +77,6 @@ The `graphqlHTTP` function accepts the following options:
 
   * **`pretty`**: If `true`, any JSON response will be pretty-printed.
 
-  * **`formatError`**: An optional function which will be used to format any
-    errors produced by fulfilling a GraphQL operation. If no function is
-    provided, GraphQL's default spec-compliant [`formatError`][] function will be used.
-
   * **`extensions`**: An optional function for adding additional metadata to the
     GraphQL response as a key-value object. The result will be added to
     `"extensions"` field in the resulting JSON. This is often a useful place to
@@ -90,6 +86,16 @@ The `graphqlHTTP` function accepts the following options:
 
   * **`validationRules`**: Optional additional validation rules queries must
     satisfy in addition to those defined by the GraphQL spec.
+
+  * **`customValidateFn`**: An optional function which will be used to validate 
+    instead of default `validate` from `graphql-js`.
+
+  * **`customFormatErrorFn`**:  An optional function which will be used to format any
+    errors produced by fulfilling a GraphQL operation. If no function is
+    provided, GraphQL's default spec-compliant [`formatError`][] function will be used.
+
+  * **`formatError`**: is deprecated and replaced by `customFormatErrorFn`. It will be 
+    removed in version 1.0.0.
 
 In addition to an object defining each option, options can also be provided as
 a function (or async function) which returns this options object. This function
@@ -288,10 +294,10 @@ graphqlHTTP.getGraphQLParams(request).then(params => {
 ## Debugging Tips
 
 During development, it's useful to get more information from errors, such as
-stack traces. Providing a function to `formatError` enables this:
+stack traces. Providing a function to `customFormatErrorFn` enables this:
 
 ```js
-formatError: error => ({
+customFormatErrorFn: error => ({
   message: error.message,
   locations: error.locations,
   stack: error.stack ? error.stack.split('\n') : [],
--- a/src/index.js
+++ b/src/index.js
@@ -174,7 +174,8 @@ function graphqlHTTP(options: Options): Middleware {
     let context;
     let params;
     let pretty;
-    let formatErrorFn;
+    let formatErrorFn = formatError;
+    let validateFn = validate;
     let extensionsFn;
     let showGraphiQL;
     let query;
@@ -217,7 +218,6 @@ function graphqlHTTP(options: Options): Middleware {
         const rootValue = optionsData.rootValue;
         const fieldResolver = optionsData.fieldResolver;
         const graphiql = optionsData.graphiql;
-        const validateFn = optionsData.customValidateFn || validate;
         context = optionsData.context || request;
 
         let validationRules = specifiedRules;
@@ -354,9 +354,7 @@ function graphqlHTTP(options: Options): Middleware {
         }
         // Format any encountered errors.
         if (result && result.errors) {
-          (result: any).errors = result.errors.map(
-            formatErrorFn || formatError,
-          );
+          (result: any).errors = result.errors.map(formatErrorFn);
         }
 
         // If allowed to show GraphiQL, present it instead of JSON.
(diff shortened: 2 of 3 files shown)


jamesmoriarty:
@IvanGoncharov aka graphql overlord can you please review?

src/index.js
```diff
@@ -463,3 +487,19 @@ function sendResponse(response: $Response, type: string, data: string): void {
...
+    if (validationCache[hash]) {
+      response.setHeader('X-Cache-Validation', 'HIT');
+
+      return validationCache[hash];
+    }
+    response.setHeader('X-Cache-Validation', 'MISS');
+
+    validationCache[hash] = op();
+
+    return validationCache[hash];
```
IvanGoncharov: There is no protection against DoS attack with a unique query per request that contains a lot of errors. Since validation errors contain AST nodes you will keep the entire query string in memory.
```js
const badQuery = `
query badQuery${counter++} {
  a b c d e #...
  a1 b1 c1 d1 e1 #...
  a2 b2 c2 d2 e2 #...
  #...
}`;
```
Since query name is different every time it means you will allocate huge amount of memory per request until your server crash.

jamesmoriarty: Hey @IvanGoncharov, Thanks for getting back to me so quickly. I really appreciate it. I wouldn't advise using an associative object in production but it was the simplest method signature without introducing something like different storage strategies. 

You can use something like https://github.com/isaacs/node-lru-cache e.g. https://repl.it/repls/MonstrousAjarArchive

```js
var LRU = require("lru-cache"),
p = console.log,
options = {max: 10},
cache = new LRU(options);

cache['test'] = 'Hello World';
p(cache['test']);
p(cache);
```

jamesmoriarty:
Hey @IvanGoncharov, the set cache property signature was a mistake. I've changed it to `set(key, value)` and `get(key`. I've changed the test to use [LruMap](http://www.collectionsjs.com/lru-map).

src/index.js
```diff
@@ -251,7 +264,24 @@ function graphqlHTTP(options: Options): Middleware {
         }
 
         // Validate AST, reporting any errors.
-        const validationErrors = validate(schema, documentAST, validationRules);
+        const string =
+          JSON.stringify(schema) +
+          JSON.stringify(source) +
+          validationRules.toString();
```
IvanGoncharov: There is a problem with your caching approach `JSON.stringify(schema)` only stringify type names and not things like fields, enum values, etc. It also doesn't take into account user-provided callbacks that affect validation e.g. `parseLiteral`
https://github.com/graphql/graphql-js/blob/8c96dc8276f2de27b8af9ffbd71a4597d483523f/src/validation/rules/ValuesOfCorrectType.js#L186

Hashing `validationRules` is affected by the exact same problems.

I think that it's impossible to implement secure & effective validation caching for all use cases.
For example, in most cases, you schema and validation rules stay the same for all requests so converting them to string and hashing is an overhead.

I think the simplest and the most flexible solution would be to allow wrapping/replacing `validate` function, similar to #391 proposal but for `validate`.

@jamesmoriarty What do you think?

jamesmoriarty: I think you're right @IvanGoncharov. We were concerned we had unsupported edge cases. We're only using a subset of graphql's feature so nothing was obviously broken. I'll see if I can invert the cache implementation with the wrapper function.

ivome:
I also think wrapping / replacing the validate function would be better. I wrote a validation rule to analyze the query complexity and reject complex queries that depends on the input variables: [graphql-query-complexity](https://github.com/slicknode/graphql-query-complexity)
To use something like this with validation caching we would have to implement some custom logic. 

Maybe we could even add a plugin system to `express-graphql` to have a well defined interface for things like this and even more complex extensions? This could also be used for #391. 

Initialization could look something like this: 

```javascript
app.get('/graphql', graphqlHTTP({
  schema: MyGraphQLSchema,
  graphiql: true,
  plugins: [
    new ValidationCachePlugin(options),
    // Other plugins
    new FakeResultPlugin(),

    // We could also add hooks for things like...
    new ProfilingPlugin(),
    // etc.
  ]
}));
```

Might be overkill for just this one feature and this would definitely need some more detailed conception but I just wanted to throw this out here to see what you think...

src/index.js
```diff
@@ -87,15 +86,14 @@ export type OptionsData = {
...
+   * An optional function to wrap validation operations.
    */
-  validationCache?: {|
-    set: (
-      key: string,
-      value: $ReadOnlyArray<GraphQLError>,
-    ) => $ReadOnlyArray<GraphQLError>,
-    get: (key: string) => $ReadOnlyArray<GraphQLError>,
-  |},
+  validationWrapper?: ?(
```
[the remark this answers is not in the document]
IvanGoncharov: @jamesmoriarty Thanks for the effort 👍 

> Unsure about the best method signature. Should it pass in the response? Would you want the same for execute?

Looking more into the source code I notice we already use the same approach for `formatError`:
https://github.com/graphql/express-graphql/blob/7ab2bc958a2184fe35fb6cd137f369b5b3877dec/src/index.js#L75-L80
https://github.com/graphql/express-graphql/blob/7ab2bc958a2184fe35fb6cd137f369b5b3877dec/src/index.js#L336-L338

I think we should duplicate this approach for both `validate` and `execute`.
Since `graphql` is a peer dependency you need to add it into your `package.json` anyway so it mean you can just write:
```js
const { validate } = require('graphql');

app.post('/graphql', graphqlHTTP({
  schema: MyGraphQLSchema,
  graphiql: false
  validate(schema, documentAST, validationRules) {
    // some caching logic
    cache[hash] = validate(schema, documentAST, validationRules);
    // more caching logic
    return cache[hash]
  },
}));
```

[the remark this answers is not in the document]
jamesmoriarty: Looks great. Will refactor towards this.

jamesmoriarty:
Thanks for the feedback @IvanGoncharov. Using the same pattern as the `formatError`worked great.

IvanGoncharov:
@jamesmoriarty Thanks 👍 
Can you also cleanup test cases (it doesn't make sense to introduce new dependencies just for tests) and make it more like `formatError`/#391?

jamesmoriarty:
Removed unnecessary dependencies and cleaned up the tests. Looks consistent with formatError and your execute PR.

Screen Shot 2019-04-03 at 9 58 31 pm

IvanGoncharov:
@jamesmoriarty Thanks a lot 👍 I will try to merge it in the next few days.

One issue I still have is with both `validate` and `execute` is that they are not very intuitive names for the functionality they enable.
Maybe `replaceValidate`/`replaceExecute`?
Would be great to hear your thoughts?

jamesmoriarty:
I agree. The existing implementation made it possible. I've had a look at the graphql-js source to get ideas on consistency. I'm thinking `customValidateFn` `customExecuteFn` and possible deprecate `formatError` and change to `customFormatErrorFn`. What do you think @IvanGoncharov?

Based on:

Screen Shot 2019-04-04 at 9 21 56 am

IvanGoncharov:
@jamesmoriarty Great suggestion 👍 I like it.
I will update #391 to use `customExecuteFn`.

> I'm thinking customValidateFn customExecuteFn and possible deprecate formatError and change to customFormatErrorFn

Make sense 👍
I think we should support deprecated `formatError` until `1.0.0` release.
Do you want to work on PR for this?

jamesmoriarty:
Sure thing.

jamesmoriarty:
Made the change to the naming, deprecating `formatError` and updating the README. I imagine there is a better way to notify the user with the deprecation warning than using `console.error`. Any feedback appreciated.

src/index.js
```diff
@@ -380,7 +401,14 @@ function graphqlHTTP(options: Options): Middleware {
...
 
-        formatErrorFn = optionsData.formatError;
+        if (optionsData.formatError) {
+          console.error(
+            '`formatError` is deprecated and replaced by `customFormatErrorFn`. It will be removed in version 1.0.0.',
+          );
+        }
+
+        formatErrorFn =
+          optionsData.customFormatErrorFn || optionsData.formatError;
```
jamesmoriarty: @IvanGoncharov interestingly I couldn't change this to:

```
optionsData.customFormatErrorFn || optionsData.formatError || formatError
// ...
.map(errorFormatFn) 
```

Without breaking the tests. It appears the 

```
.map(errorFormatFn || formatError) 
```

may be called before the options resolve. maybe around schema validation. 

Might be worth resolving outside this change.

README.md
```diff
@@ -91,6 +87,13 @@ The `graphqlHTTP` function accepts the following options:
   * **`validationRules`**: Optional additional validation rules queries must
     satisfy in addition to those defined by the GraphQL spec.
 
+  * **`customValidateFn`**: function which will be used to validate instead of default `validate` from 
+    `graphql-js`.
+
+  * **`customFormatErrorFn`**: An optional function which will be used to format any
+    errors produced by fulfilling a GraphQL operation. If no function is
+    provided, GraphQL's default spec-compliant [`formatError`][] function will be used.
```
IvanGoncharov: It should mention deprecated `formatError` and say it will be removed in `1.0.0`.

jamesmoriarty: Good point. I've added below.

jamesmoriarty:
Added to README. Thanks for spending time on this @IvanGoncharov. Much appreciated.

IvanGoncharov:
@jamesmoriarty I made a few small fixes like replacing `console.error` with `console.warn` and fixing ESLint warnings.

Issue #483, opened 2019-04-30
Support custom document loading


williambailey:
It would be nice to be able to provide a function to support custom document loading. Being able to extend/augment/replace the `new Source(...)`, `parse(source)`, `validateFn(...)` process would make implementing functionality like persisted document storage easier and also allow the server to skip the document parsing and validation stages for known/frequent documents.

IvanGoncharov:
@williambailey Great suggestion 👍 
I think the ability to replace `parse` is a great feature.
If you want to work on PR it should be very similar to #391

> extend/augment/replace the new Source(...),

`Source` is a very primitive class so I don't think we should bother to replace it.
You can always access the original query as `source.body`.
https://github.com/graphql/graphql-js/blob/cd80f94edc584c5cf3258d27b6f91ab3b72f0cce/src/language/source.js#L31-L44

IvanGoncharov:
I'm closing this issue since progress is tracked in #484

PR #484, opened 2019-05-08
Add support for customParseFn option

williambailey:
Attempt to address #483.

The current implementation in this PR provides support for synchronous `customParseFn` functions.

Eventually I think that it would be worthwhile also supporting asynchronous functions. i.e. have the `customParseFn` return type be a (let's see if i get the Flow type correct) `?Promise<DocumentNode>`.  Unfortunitely I couldn't quite get the promise handling `.then().catch()` logic right at this point in time... even when requiring the `customParseFn` be async (vs. maybe async).

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -272,7 +272,7 @@ function graphqlHTTP(options: Options): Middleware {
         }
 
         //  GraphQL source.
-        const source = new Source(query);
+        const source = new Source(query, 'GraphQL request');
 
         // Parse source to AST, reporting any syntax error.
         try {
--- a/src/index.js
+++ b/src/index.js
@@ -108,7 +108,7 @@ export type OptionsData = {
    * An optional function which will be used to create a document instead of
    * the default `parse` from `graphql-js`.
    */
-  customParseFn?: ?(params: GraphQLParams) => DocumentNode,
+  customParseFn?: ?(source: Source) => DocumentNode,
 
   /**
    * `formatError` is deprecated and replaced by `customFormatErrorFn`. It will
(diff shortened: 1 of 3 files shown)


IvanGoncharov:
@williambailey Thanks for PR 👍 

> Eventually I think that it would be worthwhile also supporting asynchronous functions. i.e. have the customParseFn return type be a (let's see if i get the Flow type correct) ?Promise<DocumentNode>. Unfortunitely I couldn't quite get the promise handling .then().catch() logic right at this point in time... even when requiring the customParseFn be async (vs. maybe async).

You don't need async since options callback can return promise of options:
```js
graphqlHTTP(graphqlHTTPOptions);

function graphqlHTTPOptions(request, response, params) {
  return getASTFromCache(params.query).then(ast => {
    schema,
    customParseFn() {
      return ast;
    }
  })
}
```
Every time you add `.then` it forces the node to switch to the next task in even loop and this adds performance penalty so it makes sense to batch all async jobs in options callback.
Also, it makes source code easier to read and reason about.

Note: you also get `params` inside options callback so it doesn't make sense to change the signature of `parse` from `graphql-js`. So can you please change `customParseFn` to:
```
customParseFn?: ?(source: Source) => DocumentNode,
```

williambailey:
Completely forgot that you could return a promise of options. PR updated.

src/index.js
```diff
@@ -265,11 +272,11 @@ function graphqlHTTP(options: Options): Middleware {
         }
 
         //  GraphQL source.
-        const source = new Source(query, 'GraphQL request');
+        const source = new Source(query);
```
IvanGoncharov: @williambailey Can you please revert this line?

IvanGoncharov:
@williambailey Thanks 👍 
Merged 🎉 
I will try to release NPM 📦 next week.

IvanGoncharov:
@williambailey Sorry for delay. I just published `0.9.0` 📦

PR #507, opened 2019-05-17
Auto prettify

rybon:
Dependent on https://github.com/graphql/graphiql/pull/830

Extends the `graphiql` option to alternatively accept an object instead of a boolean. This object has a field called `autoPrettify` that when set to `true` will pass the `autoPrettify` prop to the GraphiQL component, that will in turn run the prettify method when it mounts. This means stripped GraphQL queries will automatically be pretty printed when opened in the GraphiQL editor.

### DIFF
--- a/README.md
+++ b/README.md
@@ -66,7 +66,7 @@ The `graphqlHTTP` function accepts the following options:
   * **`graphiql`**: If `true`, presents [GraphiQL][] when the GraphQL endpoint is
     loaded in a browser. We recommend that you set
     `graphiql` to `true` when your app is in development, because it's
-    quite useful. You may or may not want it in production.
+    quite useful. You may or may not want it in production. Alternatively, instead of `true` you can pass in an options object that currently supports a single field called `autoPrettify`. When setting that field to `true`, the GraphQL queries are automatically pretty printed when opened in GraphiQL.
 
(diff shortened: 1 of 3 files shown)


rybon:
@IvanGoncharov please review.

IvanGoncharov:
@rybon I have some minor review comments but the main issue here if GraphiQL maintainers will accept this fix or not. 
If it's merged into GraphiQL I would happily help you to get this PR in shape and merge it.

rybon:
Yes, I agree. Lets wait and see what they think. Thanks for your feedback!

IvanGoncharov:
@rybon I liked your approach in this PR so I borrowed some ideas in #508.

rybon:
Cool!

rybon:
@IvanGoncharov following on the discussion in https://github.com/graphql/graphiql/pull/830, would it be possible to extend the options object with a callback field that can be called on every query, so we can make use of the `defaultQuery` prop in GraphiQL? Something like: `transformDefaultQuery: (query) => print(parse(queryFromUrlBar))`, where we pass the result to the `defaultQuery` prop?

I'm curious what you think. I think I can make a PR that would look very similar to this one based on this approach.

IvanGoncharov:
@rybon I'm not sure about this proposal I need to think about it.

PR #509, opened 2019-05-22
auto prettify

rybon:
Based on #508 this time.

### DIFF
--- a/README.md
+++ b/README.md
@@ -71,6 +71,8 @@ The `graphqlHTTP` function accepts the following options:
       * **`defaultQuery`**: An optional GraphQL string to use when no query
         is provided and no stored query exists from a previous session.
         If undefined is provided, GraphiQL will use its own default query.
+      * **`autoPrettify`**: An optional boolean that when set to true causes the
+      query to be pretty printed before being displayed.
 
   * **`rootValue`**: A value to pass as the `rootValue` to the `graphql()`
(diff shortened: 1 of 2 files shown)


rybon:
@IvanGoncharov how about this approach?

IvanGoncharov:
@rybon It feels more like a workaround. I don't want to maintain GraphiQL customization especially since GraphiQL maintainers planning big redesign effort for `1.0.0`.

To better understand your use case can you please provide more details. Is it public or private API? Why having `Prettify` button is not enough in your case?

rybon:
This does the same thing as the `Prettify` button, only automatically if `autoPrettify` is enabled in the `options` object. It does not require any changes to GraphiQL in this case. It makes use of the existing public API of GraphiQL.
It only affects the public API of this project on an opt-in basis.

IvanGoncharov:
> Basically, this feature is a small ergonomic improvement to perform that click for you.

How about adding a keyboard shortcut?
I think it will be supported by GraphiQL maintainers.

rybon:
There is a keyboard shortcut. But this happens server-side, before the page is loaded so just opening the URL is enough.

IvanGoncharov:
@rybon Thanks for the PR but I don't think `express-graphql` should be in the business of modifying GraphiQL behavior.

I'm open to generic things like passing options or listing additional GraphiQL plugins (after GraphiQL `1.0.0`) but I don't think the `express-graphql` should a place where you customize GraphiQL.

PR #516, opened 2019-06-27
Fix flow type error

HsuTing:
`mixed` had changed, after [flow@0.98.0](https://github.com/facebook/flow/releases/tag/v0.98.0):

- `mixed` refined to an array produces a read-only array.
- `mixed` refined to an object produces a read-only object.

### DIFF
--- a/package.json
+++ b/package.json
@@ -72,7 +72,7 @@
     "eslint-plugin-flowtype": "3.9.0",
     "eslint-plugin-prettier": "3.1.0",
     "express": "4.17.0",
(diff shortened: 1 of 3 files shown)


IvanGoncharov:
@HsuTing Thanks for reporting it and for detailed investigation 👍 
I fixed it in #518 by making query variables read-only both in `express-graphql` and `graphql-js`.

HsuTing:
@IvanGoncharov I'm glad to hear that. Thank for your help.

PR #565, opened 2019-11-08
Fix GraphiQL font rendering

sjparsons:
The GraphiQL page-level font-family CSS rules are not applying because the `renderGraphiQL.js` page uses a JS-type comment in a `<style>` block.

This PR is a quick fix to change to use the CSS style comment `/* ... */` so that
the first CSS rule doesn't get skipped.

Before

Screen Shot 2019-11-08 at 12 15 21 PM

After

Screen Shot 2019-11-08 at 12 15 46 PM

### DIFF
--- a/src/renderGraphiQL.js
+++ b/src/renderGraphiQL.js
@@ -69,7 +69,7 @@ add "&raw" to the end of the URL within a browser.
     }
   </style>
   <style>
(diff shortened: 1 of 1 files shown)


IvanGoncharov:
@sjparsons Thanks a lot 👍 
Merged 🎉

PR #576, opened 2019-11-28
feat(server fragments): server predefined fragments

### LINKED ISSUE #575: feat: add fragments server side
Allow clients that send queries to use predefined fragments on the server side.
The goal:
1. Shorten the query for every client by avoiding to repeat the fragment deceleration.  
2.  Allow the client to have dynamic fragments that are updated server side. For example, if a field was added to the schema server side, the fragment will update accordingly.

Ariel-Dayan:
added function addedServerFragments that concats relevent
fragments from the server to req query.
added function findFragments that return relevant fragments
for current request.
added function sliceFirstWord.
added serverFragments option.

resolves #575

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -238,6 +238,8 @@ function graphqlHTTP(options: Options): Middleware {
         const typeResolver = optionsData.typeResolver;
         const validationRules = optionsData.validationRules || [];
         const graphiql = optionsData.graphiql;
+        const serverFragments = optionsData.serverFragments;
+
         context = optionsData.context || request;
 
         // GraphQL HTTP only supports GET and POST methods.
@@ -247,7 +249,7 @@ function graphqlHTTP(options: Options): Middleware {
         }
 
         // Get GraphQL params from the request and POST body data.
-        query = params.query;
+        query = serverFragments ? addServerFragments(serverFragments, params.query) : params.query;
         variables = params.variables;
         operationName = params.operationName;
         showGraphiQL = canDisplayGraphiQL(request, params) && graphiql;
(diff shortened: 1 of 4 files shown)


acao:
nice! so we have a standard for this using graphql-config that is used pretty widely for loading additional fragments, for example with playground or graphiql. maybe we can build off of that so folks wont have to maintain duplicate configuration for this and the likely soon to be standard graphql-config?

src/index.js
```diff
@@ -509,3 +516,87 @@ function sendResponse(response: $Response, type: string, data: string): void {
...
+/**
+ * Helper recursive function that finds all the fragments from the server that are used in the current request.
+ * 
+ * @param   { string }      serverFragments - Fragments from the server.
+ * @param   { string }      query           - Query from the request.
+ * @param   { Set<string> } fragmentsInUsed - Set of relevant fragments.
+ * 
+ * @returns { Set<string> }                 - relevant fragments for current request.
+ */
+function findFragments(serverFragments: string, query: string, fragmentsInUsed: Set<string>): Set<string> {
```
acao: it would be much simpler and more concise to use the `graphql` `visit()` function here I think

src/index.js
```diff
@@ -509,3 +516,87 @@ function sendResponse(response: $Response, type: string, data: string): void {
...
+ */
+function addServerFragments(serverFragments: string, query: string): string {
+  let fragmentsInUsed = '';
+
+  [...(findFragments(serverFragments, query, new Set()))].forEach(fullFragment => {
+    fragmentsInUsed += `fragment ${fullFragment} `;
+  })
+
+  return `${fragmentsInUsed}\n${query}`;
+}
```
acao: EOL issue

src/index.js
```diff
@@ -509,3 +516,87 @@ function sendResponse(response: $Response, type: string, data: string): void {
...
+ * @param {*} serverFragments - Fragments from the server.
+ * @param {*} query           - Query from the request.
+ * 
+ * @returns                   - Concat relevant fragments to request query
+ */
+function addServerFragments(serverFragments: string, query: string): string {
+  let fragmentsInUsed = '';
+
+  [...(findFragments(serverFragments, query, new Set()))].forEach(fullFragment => {
+    fragmentsInUsed += `fragment ${fullFragment} `;
```
acao: itll probably be easier and more perfomant to generate this from AST as well

src/index.js
acao: https://graphql.org/graphql-js/language/#visit just parse the operation string to AST, and visit all the `FragmentSpreadDefinition` node types i think it would be for this case.

```js
import { visit, print } from 'graphql'
const operationAST = parse(query)
const serverFragments  = ''
visit(operationAst, {
  FragmentDefinition(node) {
    // do matching/filtering logic here using full ast node.
    // node.name.value, etc
    serverFragments += print(node)
  }
})
```

acao:
also https://travis-ci.org/graphql/express-graphql/jobs/618258347#L236

this is looking good though! i'll gladly help you get it ready for @IvanGoncharov to give this a final pass

as per my earlier comment re: graphql-config, it will make sense as an abstraction around this interface.

src/index.js
acao: I could be wrong though, about which option is more performant, as parsing each query and then re-printing the nodes entails another expense, though much of this could be memoized, eventually cached even?

IvanGoncharov:
@Ariel-Dayan @acao Don't want to be a party killer but I don't think it's the right feature for the reference implementation of GraphQL Server.
By supporting incomplete queries (in this case without fragments) we force the entire ecosystem (client tools, graphql gateways, etc.) to do the same.

> 1. Shorten the query for every client by avoiding to repeat the fragment declaration.

This is solved by persistent queries

> Allow the client to have dynamic fragments that are updated server side. For example, if a field was added to the schema server side, the fragment will update accordingly.

It brake assumption for client-side tools they since they were compiled/deployed with different sets of fields and this invalidates GraphQL type-safety/"predictable payload" principle.

At the same time if you want you can add this feature quite easily in your code without modifying `express-graphql`:
```js
import { parse } from 'graphql';
// ...
app.use(
  '/graphql',
  graphqlHTTP({
    schema: MySessionAwareGraphQLSchema,
    graphiql: true,
    customParseFn(source) {
      const ast = parse(source);
      // code that inject fragments into ast
      const serverFragmentString  = ''
      visit(ast, {
        FragmentSpread(node) {
          const name = node.name.value;
          if(serverFragments.includes(name)) {
            serverFragmentString += serverFragments.get(name);
          }
        }
      });
      
      return concatAST(ast, parse(serverFragmentString));
    }
  }),
);
```

PR #578, opened 2019-12-04
Move graphql dependency to correct section

luke88jones:
The graphql dependency is currently under the devDependencies section so causes missing module errors when included in a project.

### DIFF
--- a/package.json
+++ b/package.json
@@ -44,6 +44,7 @@
   "dependencies": {
     "accepts": "^1.3.7",
     "content-type": "^1.0.4",
+    "graphql": "14.5.8",
     "http-errors": "^1.7.3",
     "raw-body": "^2.4.1"
   },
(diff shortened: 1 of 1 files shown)


acao:
@luke88jones thanks so much for submitting this!

`graphql-express` specifies `graphql` as a peer dependency, so you should install whichever version that meets the resolution criteria. So you'll want to add to your project's dependencies any release after `14.4.1` and before `15.0.0-alpha.x`. Peer deps allow folks to manage the underlying version themselves, thus express-graphql doesnt need a release for every release of `graphql`.

here's a more in depth explanation of how to resolve `peerDependencies` issues you might be having:
https://dev.to/btdev/what-are-those-peerdepenedencies-in-a-nodejs-project-51jo

hope that helps!
https://dev.to/btdev/what-are-those-peerdepenedencies-in-a-nodejs-project-51jo

PR #583, opened 2020-01-09
Support for experimental @defer & @stream

robrichard:
Support for streaming multi-part http responses as implemented by 

https://github.com/graphql/graphql-js/pull/2319
https://github.com/relay-tools/fetch-multipart-graphql

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -275,7 +275,7 @@ export function graphqlHTTP(options: Options): Middleware {
       // https://graphql.github.io/graphql-spec/#sec-Response-Format
       if (optionsData.extensions) {
         extensionsFn = (payload: AsyncExecutionResult) => {
-          /* istanbul ignore next condition not reachable, required for flow */
+          /* istanbul ignore else: condition not reachable, required for typescript */
           if (optionsData.extensions) {
             return optionsData.extensions({
               document: documentAST,
@@ -285,6 +285,8 @@ export function graphqlHTTP(options: Options): Middleware {
               context,
             });
           }
+          /* istanbul ignore next: condition not reachable, required for typescript */
+          return undefined;
         };
       }
 
@@ -362,21 +364,23 @@ export function graphqlHTTP(options: Options): Middleware {
           fieldResolver,
           typeResolver,
         });
+
+        if (isAsyncIterable(executeResult)) {
+          // Get first payload from AsyncIterator. http status will reflect status
+          // of this payload.
+          const asyncIterator = getAsyncIterator<ExecutionResult>(
+            executeResult,
+          );
+          const { value } = await asyncIterator.next();
+          result = value;
+        } else {
+          result = executeResult;
+        }
       } catch (contextError: unknown) {
         // Return 400: Bad Request if any execution context errors exist.
         throw httpError(400, 'GraphQL execution context error.', {
           graphqlErrors: [contextError],
         });
-      }
-
-      if (isAsyncIterable(executeResult)) {
-        // Get first payload from AsyncIterator. http status will reflect status
-        // of this payload.
-        const asyncIterator = getAsyncIterator<ExecutionResult>(executeResult);
-        const { value } = await asyncIterator.next();
-        result = value;
-      } else {
-        result = executeResult;
       }
 
       if (extensionsFn) {
@@ -412,9 +416,12 @@ export function graphqlHTTP(options: Options): Middleware {
           undefined,
           error,
         );
-        result = { data: undefined, errors: [graphqlError] };
+        executeResult = result = { data: undefined, errors: [graphqlError] };
       } else {
-        result = { data: undefined, errors: error.graphqlErrors };
+        executeResult = result = {
+          data: undefined,
+          errors: error.graphqlErrors,
+        };
       }
     }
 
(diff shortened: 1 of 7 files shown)


acao:
looks great, thanks @robrichard! once we get one of the graphql-js proposals merged for 15.0.0 we can revisit the tests here. saw there are two proposal PRs for graphql-js

src/index.ts
```diff
@@ -388,6 +418,31 @@ export function graphqlHTTP(options: Options): Middleware {
       errors: result.errors?.map(formatErrorFn),
     };
 
+    if (isAsyncIterable(executeResult)) {
+      response.setHeader('Content-Type', 'multipart/mixed; boundary="-"');
+      sendPartialResponse(pretty, response, formattedResult);
+      for await (let payload of executeResult) {
```
danielrearden, line 424: @robrichard What happens when the request is terminated by the client (i.e. I navigate away from the page or close the browser)? Would it make sense to add an event listener to response for the `close` event and call the AsyncIterator's `return` method inside?

robrichard: Yes, I think that makes sense. I will implement that.

n1ru4l:
I got a bit too excited so I created a playground making it easier for people to test this new stuff! https://github.com/n1ru4l/graphql-bleeding-edge-playground

Awesome work!!!

acao:
@n1ru4l awesome! it looks great in graphiql as well. we are going to archive playground soon though :(

n1ru4l:
ah wait it is graphiql 😂 playground ist just the name of the repo it uses graphiql :D

acao:
@n1ru4l oh awesome I see, thanks for that! did you see this? https://github.com/graphql/graphiql/discussions/1700#discussioncomment-113214

acao:
@n1ru4l are you interested in helping me create this PR for the new fetcher patterns? Otherwise I might not have time until next weekend :/

n1ru4l:
@robrichard Would it be possible for you to do single commit instead of force-pushing, I am having a hard time following the changes 😅

robrichard:
I've been force pushing because it's harder to keep rebasing when there are a lot of commits and this PR has been open for a long time.

src/index.ts
```diff
@@ -344,16 +372,18 @@ export function graphqlHTTP(options: Options): Middleware {
         });
       }
 
-      // Collect and apply any metadata extensions if a function was provided.
-      // https://graphql.github.io/graphql-spec/#sec-Response-Format
+      if (isAsyncIterable(executeResult)) {
```
n1ru4l, line 375: Since executeFn is typed to return an `AsyncIterableIterator`, it would make sense to instead use a `isAsyncIterableIterator` function. That way it is also possible to call `executeResult.return?.()` in case the request is aborted without adding ts-ignores.

n1ru4l:
It would also make sense to adjust the `renderGraphiQL` function to provide a multipart compatible fetcher

src/index.ts
n1ru4l: Nevermind, if you use the isAsyncIterable function th type checker already is aware of the actual type (`AsyncIterableIterator`), making it easy to call .return()

src/index.ts
```diff
@@ -344,16 +372,18 @@ export function graphqlHTTP(options: Options): Middleware {
...
+        // Get first payload from AsyncIterator. http status will reflect status
+        // of this payload.
+        const asyncIterator = getAsyncIterator<ExecutionResult>(executeResult);
+        const { value } = await asyncIterator.next();
```
IvanGoncharov, line 379: `await` can throw and should be inside `try` 🔼

IvanGoncharov: test for that would be great. You can do that by using custom `executeFn` that return an async iterator.

robrichard: @IvanGoncharov do you think it makes sense for the `try/catch` that is currently wrapping `await executeFn({...})` to be expanded to also wrap this?

IvanGoncharov: @robrichard Yes, I think it would work 👍

src/index.ts
```diff
@@ -523,6 +604,36 @@ function canDisplayGraphiQL(request: Request, params: GraphQLParams): boolean {
...
+/**
+ * Helper function for sending part of a multi-part response using only the core Node server APIs.
+ */
+function sendPartialResponse(
+  pretty: boolean,
+  response: Response,
+  result: FormattedExecutionResult | FormattedExecutionPatchResult,
+): void {
+  const json = JSON.stringify(result, null, pretty ? 2 : 0);
+  const chunk = Buffer.from(json, 'utf8');
```
maraisr, line 616: will this always be utf8?

src/index.ts
```diff
@@ -523,6 +604,36 @@ function canDisplayGraphiQL(request: Request, params: GraphQLParams): boolean {
...
+  const data = [
+    '',
+    '---',
+    'Content-Type: application/json; charset=utf-8',
```
maraisr, line 620: as per my other comment, can we assume the consumer will send utf8?

maraisr:
Wish github had stacked PR's 😔 https://github.com/robrichard/express-graphql/pull/1

src/index.ts
acao: good question, I'm fairly sure that `utf8` is the only way that `application/json` content can be encoded

acao: this has me wanting to bring up protocolbuffers, messagepack and other binary formats, as well as yaml, however that seems way out of scope for this PR, and should probably begin as an HTTP transport RFC proposal. GraphiQL users have on multiple occasions requested support for these formats, especially for server-to-server transmissions and iot

back to this discussion, though, would it work to use a binary encoding format `latin1` with JSON? I've never tried that myself. if someone was returning a binary format like BSON for example, then we should be returning content type of `application/bson`, right?  Is there a case where someone would be sending `application/json`?

if there are indeed more ways to mix and match charset and content type with JSON than I'm aware of, maybe this should be controlled by the request headers then, and fall back to `utf8` as default?

acao: Worth noting that the transport RFC spec doesn't require `utf8` necessarily (yet):
https://github.com/graphql/graphql-over-http/blob/master/rfcs/IncrementalDelivery.md

and the reference client implementation expects `utf8`:
https://github.com/relay-tools/fetch-multipart-graphql/blob/dfa54d0c649ecbbedac3e9764998f63894367505/src/fetch.js#L26

maraisr: meros always at this time uses UTF8. It's a force of habit. But there might be people out there needing some out of the ordinary encodings. 

Maybe it's a cross that bridge when we get there?

robrichard: express-graphql already only sends utf8 so I don't think there's any change in behavior here: https://github.com/graphql/express-graphql/blob/master/src/index.ts#L531

As far as the spec, I'm not sure if it's worth specifying what encodings are required. FWIW it seems like JSON requires either UTF-8, UTF-16, or UTF-32 but recommends UTF8 for maximum interoperablity.

https://tools.ietf.org/html/rfc7159#section-8.1

src/index.ts
robrichard: I added a handler for the close event to call `return` on the underlying async iterable. 

@IvanGoncharov this required disabling some lint rules (see https://github.com/graphql/express-graphql/pull/583/commits/d883ae71eeacb9b148f32d01f68b20c831ceca07#diff-a2a171449d862fe29692ce031981047d7ab755ae7f84c707aef80701b3ea0c80R419). Do you think this is ok or do you have a better way to implement this?

PR #586, opened 2020-02-22
feat: upgrade graphiql to 0.17.5

acao:
bumping the graphiql implementation from 0.14.2 to 0.17.5.
lockfile includes requisite bumps to dependencies

(update: remove node 8 from travis config)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,7 +7,6 @@ cache: yarn
 node_js:
   - '12'
   - '10'
-  - '8'
 
 script: |
   if [[ "$(node -pe process.version)" == v10.* ]]; then # Is latest LTS?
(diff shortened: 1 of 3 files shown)


acao:
ah so graphiql requires node 9.7 or greater

https://blog.risingstack.com/update-nodejs-8-end-of-life-no-support/
node 8 is no longer supported this year at all. so I'll update the PR to bump the version in travis, or just create another PR?

danielrearden:
Should the `engines` property in package.json also be updated to reflect dropping support for node 8?

acao:
well, the engine dependency is now not strict which makes this easier

it can just inherit the warning from graphiql itself, but i guess we could add the engines value here too. just no strict

danielrearden:
@acao @IvanGoncharov With #614 merged, we could close this but there's still the question of removing Node 8 from the Travis config

IvanGoncharov:
@danielrearden Thanks for the ping 👍 
I plan to sync this repo as much as possible with `graphql-js` including GitHub actions set up so I will do that today and it has only currently supported node versions in it.

PR #591, opened 2020-03-19
TS: Allow to pass object as graphiql option

juffalow:
When using typescript, it is not allowed to pass an object as graphiql value. Only boolean values are allowed. This disables the user to set his own defaultQuery that is shown by default on GraphiQL render.


juffalow:
Workaround for now:

```typescript
app.use('/graphql', graphqlHTTP({
  context,
  graphiql: <boolean> <unknown> {
    defaultQuery: config.defaultQuery,
  },
  schema,
}));
```

ppsirius:
Anyone could merge this?

PR #594, opened 2020-04-06
Fix customExecuteFn return type

felixputera:
Make `customExecuteFn` returns `ExecutionResult | Promise<ExecutionResult>` similar to [`graphql.execute`](https://github.com/graphql/graphql-js/blob/278bde0a5cd71008452b555065f19dcd1160270a/src/execution/execute.js#L141-L144).


felixputera:
@IvanGoncharov sorry for mentioning you directly like this. Can I ask for your help to review this tiny PR? Thanks in advance!

PR #622, opened 2020-06-07
Sync TS typings with Flow typings

danielrearden:
This PR only changes the existing TS types.

### DIFF
--- a/types/index.d.ts
+++ b/types/index.d.ts
@@ -19,7 +19,7 @@ import {
 export = graphqlHTTP;
 
 type Request = IncomingMessage;
-type Response = ServerResponse;
+type Response = ServerResponse & { json?: (data: unknown) => void };
 
 declare namespace graphqlHTTP {
   /**
@@ -44,7 +44,7 @@ declare namespace graphqlHTTP {
      * query exists from a previous session.  If undefined is provided, GraphiQL
      * will use its own default query.
      */
-    defaultQuery: string;
+    defaultQuery?: string;
   }
 
   interface OptionsData {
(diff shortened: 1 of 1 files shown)


danielrearden:
@IvanGoncharov My current plan is to:

* Fix existing types (this PR)
* Break up the type declarations and move them into `/src` (this will also convert from using a namespace to ES module syntax)
* Add any missing type declarations
* Convert the tests to TS
* Convert the rest of the files to TS, extracting the Flow typings

types/index.d.ts
```diff
@@ -56,64 +58,64 @@ declare namespace graphqlHTTP {
     /**
      * A value to pass as the context to the graphql() function.
      */
-    context?: unknown;
+    context?: Maybe<unknown>;
```
IvanGoncharov, line 61: `unknown` should include `null` and `undefined` right?

types/index.d.ts
```diff
@@ -56,64 +58,64 @@ declare namespace graphqlHTTP {
...
 
     /**
      * An object to pass as the rootValue to the graphql() function.
      */
-    rootValue?: unknown;
+    rootValue?: Maybe<unknown>;
```
IvanGoncharov, line 66: same

types/index.d.ts
IvanGoncharov: I see it marked as `?mixed` in Flow.
So this change is ok 👍

types/index.d.ts
IvanGoncharov: 👍

types/index.d.ts
IvanGoncharov: I think it can fixed in separate PR both in Flow and TS

types/index.d.ts
```diff
@@ -56,64 +58,64 @@ declare namespace graphqlHTTP {
...
 
     /**
      * A boolean to configure whether the output should be pretty-printed.
      */
-    pretty?: boolean | null;
+    pretty?: Maybe<boolean>;
```
IvanGoncharov, line 71: One thing to note: after TS conversion we need to get rid of `Maybe` types, AFAIK they are not part of idiomatic TS and at that point would be just a Flow legacy.

IvanGoncharov:
@danielrearden It looks like TS by default supports `undefined` as values of optional arguments and fields: https://www.typescriptlang.org/play/index.html#code/C4TwDgpgBA8mwEsD2A7AzlAvFA3gKCigDMB+ALijWACcEUBzAbjwF9m8BjVKqJeNAIwU4ibllwtO3YL34AmYfGTpxOYhQCuKACYQidCNqiSu6GX2BoAzItErsaohRQaANq+NSzsywBZbyhgO6lAA5ERISKHGQA

Moreover, it looks like Flow has the same behaviour:
https://flow.org/try/#0C4TwDgpgBA8mwEsD2A7AzlAvFA3gKCigDMB+ALijWACcEUBzAGjwF8BuPPAY1SqiXhoAjBTiJeWXC269g-QQCZR8ZOkk5iFAK4oAJhCJ0IuqNJ7o5A4GgAsy8WuwaiFAOREkSV6c5EdXB2IACipqckoaOnoASik8IiCdfUMUY2j4oPciV3SgA

IvanGoncharov:
@danielrearden Based on the previous comment I cleaned up Flow typings in #624 
Can you please sync this PR?

> My current plan is to:

Sounds good 👍

PR #626, opened 2020-06-09
Convert CommonJS module syntax to ES6

danielrearden:


README.md
```diff
@@ -321,9 +321,9 @@ running a GraphQL request. This function is used internally to handle the
 incoming request, you may use it directly for building other similar services.
 
 ```js
-const graphqlHTTP = require('express-graphql');
+const { graphqlHTTP } = require('express-graphql');
```
IvanGoncharov, line 324:
```suggestion
const { graphqlHTTP, getGraphQLParams } = require('express-graphql');
```

types/tsconfig.json
```diff
@@ -8,6 +8,6 @@
     "strictFunctionTypes": true,
     "noEmit": true,
     "baseUrl": ".",
-    "paths": { "express-graphql": ["."] }
+    "paths": { "express-graphql": ["./index"] }
```
IvanGoncharov, line 11: Why did you change that?

types/index.d.ts
```diff
@@ -16,179 +16,174 @@ import {
   GraphQLTypeResolver,
 } from 'graphql';
 
-export = graphqlHTTP;
+export {};
```
IvanGoncharov, line 19: Why do we need this construct?

types/index.d.ts
```diff
@@ -16,179 +16,174 @@ import {
...
 
 type Request = IncomingMessage;
 type Response = ServerResponse & { json?: (data: unknown) => void };
 
-declare namespace graphqlHTTP {
+/**
+ * Used to configure the graphqlHTTP middleware by providing a schema
```
IvanGoncharov, line 25: So it just indentation changes you didn't change anything inside, right?

danielrearden: Removed the namespace and exported the individual types

types/tsconfig.json
danielrearden: Dangit, sorry missed that

types/index.d.ts
danielrearden: dtslint was complaining about the types we don't export

> All declarations in this module are exported automatically. Prefer to explicitly write 'export' for clarity. If you have a good reason not to export this declaration, add 'export {}' to the module to shut off automatic exporting. See: https://github.com/Microsoft/dtslint/blob/master/docs/strict-export-declare-modifiers.md

types/tsconfig.json
IvanGoncharov: @danielrearden No problem, it happens, e.g. yesterday I committed `console.log` into `graphql-js` 🤦

types/index.d.ts
IvanGoncharov: Wow, now I understand why it was complaining in `graphql-js`.

PR #628, opened 2020-06-10
Refactor TS declaration files

danielrearden:
This PR moves the type declarations inside `src` and splits them up by file. It also moves `tsconfig.ts` into the project root and makes some necessary changes to keep dtslint working.

### DIFF
--- a/src/__tests__/http-test.js
+++ b/src/__tests__/http-test.js
@@ -29,6 +29,15 @@ import {
 
 import { graphqlHTTP } from '../index';
 
+// TODO Improve typings after converting to TypeScript
+type Server = () => {|
+  get: (...args: Array<mixed>) => mixed,
+  post: (...args: Array<mixed>) => mixed,
+  put: (...args: Array<mixed>) => mixed,
+  request: () => any,
+  use: (...args: Array<mixed>) => any,
+|};
+
 const QueryRootType = new GraphQLObjectType({
   name: 'QueryRoot',
   fields: {
@@ -137,7 +146,7 @@ describe('GraphQL-HTTP tests for restify', () => {
   });
 });
 
-function runTests(server) {
+function runTests(server: Server) {
   describe('GET functionality', () => {
     it('allows GET with query param', async () => {
       const app = server();
--- a/tsconfig.json
+++ b/tsconfig.json
@@ -1,7 +1,9 @@
 {
   "compilerOptions": {
     "module": "commonjs",
-    "lib": ["es6", "esnext.asynciterable"],
+    "lib": ["es2018", "esnext.asynciterable"],
+    "target": "es2018",
+    "moduleResolution": "node",
     "noImplicitAny": true,
     "noImplicitThis": true,
     "strictNullChecks": true,
--- a/package.json
+++ b/package.json
@@ -32,7 +32,7 @@
     "test:ci": "npm ci && npm run prettier:check && npm run lint && npm run check && npm run testonly:cover && npm run build && npm run check:integrations",
     "testonly": "mocha src/**/__tests__/**/*.js",
     "testonly:cover": "nyc npm run testonly",
-    "lint": "eslint src types resources integrationTests",
+    "lint": "eslint src resources integrationTests",
     "prettier": "prettier --ignore-path .gitignore --write --list-different '**/*.{js,ts,md,json,yml}'",
     "prettier:check": "prettier --ignore-path .gitignore --check '**/*.{js,ts,md,json,yml}'",
     "check": "flow check",
--- a/resources/build.js
+++ b/resources/build.js
@@ -25,10 +25,11 @@ if (require.main === module) {
 
       const cjs = babelBuild(srcPath, { envName: 'cjs' });
       fs.writeFileSync(destPath, cjs);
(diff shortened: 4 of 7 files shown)


resources/check-ts.js
```diff
@@ -0,0 +1,19 @@
+// @noflow
```
danielrearden, line 1: @IvanGoncharov I couldn't figure out a way to point `dtslint` at `src` and have it look for the `tsconfig.json` in the root. Pointing it at the root also doesn't work because it explicitly looks for an `index.d.ts` file in whatever dir you set. TL;DR This is a bandaid solution to keep `tsconfig.json` in the root without duplicating it inside `src`.

Once we've converted everything to TS, we'll be able to drop `dtslint` altogether and we can remove this script.

tsconfig.json
```diff
@@ -1,13 +1,16 @@
 {
   "compilerOptions": {
     "module": "commonjs",
+    "target": "es2018",
```
danielrearden, line 4: Added some properties to align with the `tsconfig.json` added [here](https://github.com/graphql/graphql-js/pull/2609).

resources/check-ts.js
IvanGoncharov: @danielrearden We need `dtslint` in the long run to ensure that we ship correct typings that work with older versions of typescript since https://github.com/sandersn/downlevel-dts is not really mature yet.
Better solution would be to build npm package into some tmp directory and run dtslint against this temp dir.

resources/check-ts.js
```diff
@@ -0,0 +1,19 @@
...
+
+const source = 'tsconfig.json';
+const destination = path.join('./src', 'tsconfig.json');
+
+fs.copyFileSync(source, destination);
+
+try {
+  exec('dtslint src', { stdio: 'inherit' });
+} finally {
+  fs.unlinkSync(destination);
```
IvanGoncharov, line 18: Not an issue for me since I use Vim but flickering files create an issues for IDE users.

tsconfig.json
```diff
@@ -1,13 +1,16 @@
...
+    "moduleResolution": "node",
     "lib": ["es6", "esnext.asynciterable"],
```
IvanGoncharov, line 6: We can use full `es2018` here.

IvanGoncharov:
@danielrearden I need to think about `dtslint` and I will try to find correct solution in `graphql-js` so you can latter port it here.

Meanwhile, it would be great if you can ensure that we have types on all function arguments (except inline functions and functions defined as object properties) since TS will require us to define it anyway but TS conversion diff will be way smaller.
You can try to find some set of options that will force ESLint to detect such places (not really necessary here but we can use this approach for larger code base in `graphql-js`). 
Importantly don't type functions unless required by TS.

resources/check-ts.js
danielrearden: @IvanGoncharov Boo 😭  Welp, there's always monkey-patching.

On a related note, maybe you were already planning to do this, but we could [publish separate files targeting different TS versions](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-3-1.html#version-selection-with-typesversions) for the type declarations.

IvanGoncharov: > On a related note, maybe you were already planning to do this, but we could publish separate files targeting different TS versions for the type declarations.

yes, it's what the TS team wants library authors to do with `downlevel-dts`.

IvanGoncharov:
@danielrearden I merged #637 and it should unblock this PR.
Can you please rebase?

danielrearden:
@IvanGoncharov Rebased!

resources/check-ts.js
```diff
@@ -0,0 +1,30 @@
...
+const srcFiles = readdirRecursive('./src');
+for (const filepath of srcFiles) {
+  const srcPath = path.join('./src', filepath);
+  const destPath = path.join('./tmp', filepath);
+
+  if (filepath.endsWith('.ts')) {
+    fs.copyFileSync(srcPath, destPath);
+  }
+}
+
```
IvanGoncharov, line 23: @danielrearden You don't need this file anymore, right?

src/__tests__/http-test.js
```diff
@@ -29,6 +29,14 @@ import {
 
 import { graphqlHTTP } from '../index';
 
+type Server = () => {|
```
IvanGoncharov, line 32: Leave a TODO to improve typing after we switch to TS and will have typings for `supertest` and other libs.

tsconfig.json
```diff
@@ -1,7 +1,9 @@
 {
   "compilerOptions": {
     "module": "commonjs",
-    "lib": ["es6", "esnext.asynciterable"],
+    "target": "es2018",
+    "moduleResolution": "node",
+    "lib": ["es2018", "esnext.asynciterable"],
```
IvanGoncharov, line 6: nit: can you move it before `target` minimize the diff?

IvanGoncharov:
@danielrearden Left a couple of comments, everything else looks good 👍

tsconfig.json
IvanGoncharov: Merging as is but can you please research if `es2018` already includes async iteratable?
Also can you please research what the best way tie it to Node v10?

PR #629, opened 2020-06-10
Fix type for extensions option

danielrearden:


src/index.js
```diff
@@ -125,7 +125,9 @@ export type OptionsData = {|
    *
    * This function may be async.
    */
-  extensions?: (info: RequestInfo) => { [key: string]: mixed, ... },
+  extensions?: (
+    info: RequestInfo,
+  ) => { [key: string]: mixed, ... } | Promise<{ [key: string]: mixed, ... }>,
```
IvanGoncharov, line 130: @danielrearden Suggestion (to make my life easier): It would be great if you use references to code that motivated the change, in this case:
https://github.com/graphql/express-graphql/blob/3c26db785876160de1fe4c0ed867af35daccdceb/src/index.js#L351
Not essential but it will save me from figuring out where to find matching code.

PR #630, opened 2020-06-10
Add missing parameter types

danielrearden:


src/renderGraphiQL.js
```diff
@@ -19,9 +19,9 @@ export type GraphiQLOptions = {|
 |};
 
 // Ensures string values are safe to be used within a <script> tag.
-function safeSerialize(data) {
+function safeSerialize(data: mixed) {
```
IvanGoncharov, line 22:
```suggestion
function safeSerialize(data: mixed): string {
```
Also, return values on those functions don't cause any issue with Flow we can add them.

src/parseBody.js
```diff
@@ -76,7 +76,10 @@ export async function parseBody(
 const jsonObjRegex = /^[ \t\n\r]*\{/;
 
 // Read and parse a request body.
-async function readBody(req, typeInfo) {
+async function readBody(
+  req: $Request,
+  typeInfo: {| type: string, parameters: { [param: string]: string, ... } |},
```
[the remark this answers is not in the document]
IvanGoncharov: It's actually helpful to have `TODO` in code for such a thing.

src/__tests__/http-test.js
```diff
@@ -75,11 +76,13 @@ function urlString(urlParams?: { [param: string]: string, ... }): string {
   return string;
 }
 
-[
+const implementations: Array<[(...args: Array<any>) => any, string]> = [
```
IvanGoncharov, line 79: This code was a constant source of frustration for me and now it even introduces bunch of any types.
So I just rewrote it in #631 
Can you please rebase?

src/__tests__/http-test.js
```diff
@@ -1962,7 +1963,9 @@ function runTests(server) {
   });
 
   describe('Custom validation rules', () => {
-    const AlwaysInvalidRule = function (context) {
+    const AlwaysInvalidRule = function (
+      context: ValidationContext,
+    ): { [key: string]: mixed, ... } {
```
IvanGoncharov, line 1968: You can use `ASTVisitor` as return value.

src/renderGraphiQL.js
```diff
@@ -19,9 +19,9 @@ export type GraphiQLOptions = {|
 |};
 
 // Ensures string values are safe to be used within a <script> tag.
-function safeSerialize(data) {
+function safeSerialize(data: mixed): string {
```
IvanGoncharov, line 22: Actually Flow is correct and you don't need `any` if you change the argument to `data: string | null`

IvanGoncharov:
@danielrearden Left a couple of comments and everything else looks great 👍

src/renderGraphiQL.js
danielrearden: Oops, I thought we were passing other types here, not just strings. I ended up using `?string` here to match the types of the fields in `GraphiQLData`.

PR #640, opened 2020-06-16
Convert tests to TS

danielrearden:


src/__tests__/http-test.ts
```diff
@@ -96,11 +95,11 @@ describe('GraphQL-HTTP tests for connect', () => {
 
     return {
       request: () => supertest(app),
-      use: (...args) => app.use(...args),
+      use: app.use.bind(app),
```
danielrearden, line 98: Tweaked these functions to get around TS complaining about wrong number of arguments.

src/__tests__/http-test.ts
```diff
@@ -860,7 +859,7 @@ function runTests(server: Server) {
         .set('Content-Type', 'application/json')
         .set('Content-Encoding', 'gzip');
 
-      req.write(zlib.gzipSync('{ "query": "{ test }" }'));
+      await req.write(gzipSync('{ "query": "{ test }" }'));
```
danielrearden, line 862: `write` is typed as returning a Promise

src/__tests__/http-test.ts
```diff
@@ -1572,7 +1573,7 @@ function runTests(server: Server) {
       const response = await app.request().put(urlString({ query: '{test}' }));
 
       expect(response.status).to.equal(405);
-      expect(response.headers.allow).to.equal('GET, POST');
+      expect(response.get('allow')).to.equal('GET, POST');
```
danielrearden, line 1576: `headers` was a private property

.eslintrc.yml
```diff
@@ -175,7 +175,7 @@ rules:
   import/first: error
   import/exports-last: off
   import/no-duplicates: error
-  import/no-namespace: error
+  import/no-namespace: off
```
IvanGoncharov, line 178: Can you please add `FIXME` comment describing why you disable it?

package.json
```diff
@@ -80,6 +89,7 @@
     "restify": "4.3.2",
     "sinon": "9.0.2",
     "supertest": "4.0.2",
+    "ts-node": "^8.10.2",
```
IvanGoncharov, line 92: Can you please use the exact versions?

src/__tests__/usage-test.ts
```diff
@@ -99,8 +80,7 @@ describe('Useful errors when incorrectly used', () => {
   });
 
   it('validates schema before executing request', async () => {
-    // $DisableFlowOnNegativeTest
```
IvanGoncharov, line 102: Please change it to ts-expect-error instead of cast

src/__tests__/usage-test.ts
```diff
@@ -1,7 +1,5 @@
-// @flow strict
-
-import express from 'express';
-import request from 'supertest';
+import * as express from 'express';
+import * as request from 'supertest';
```
IvanGoncharov, line 2: Can you use some tsconfig.json option to allow the previous syntax?

src/__tests__/http-test.ts
```diff
@@ -1117,9 +1116,11 @@ function runTests(server: Server) {
     app.get(
       urlString(),
       graphqlHTTP((req, res) => {
+        // eslint-disable-next-line @typescript-eslint/no-unnecessary-condition
         if (req) {
           hasRequest = true;
         }
+        // eslint-disable-next-line @typescript-eslint/no-unnecessary-condition
```
IvanGoncharov, line 1123: add fixme comment to not that we want to fix it in the future.

src/__tests__/http-test.ts
IvanGoncharov: Can you please extract it into separate PR and fix it still in Flow?
Also can you please reference the `write` definition so I can see it?

src/__tests__/http-test.ts
```diff
@@ -491,8 +490,8 @@ function runTests(server: Server) {
       const app = server();
 
       // Middleware that adds req.foo to every request
-      app.use((req, _res, next) => {
-        req.foo = 'bar';
+      app.use((req: Request, _res: Response, next: Next) => {
```
IvanGoncharov, line 493: Do you need to explicitly state types here?
I thought TS will infer it from `app.use`.

src/__tests__/http-test.ts
danielrearden: @IvanGoncharov I guess this is more of a false positive.

Screen Shot 2020-06-17 at 3 46 51 PM

The result of calling `write` is a thenable and is not assigned to a variable so ESLint thinks there's a floating Promise.

Other than awaiting the code, we could also either ignore the `no-floating-promises` rule or chain the method with the preceding ones:

```
const req = app
        .request()
        .post(urlString())
        .set('Content-Type', 'application/json')
        .set('Content-Encoding', 'gzip')
        .write(gzipSync('{ "query": "{ test }" }'));
```

danielrearden: I think adding the method to the existing chain makes the most sense. Off-hand, I'm not sure why that wasn't done like that in the first place.

src/__tests__/http-test.ts
```diff
@@ -13,8 +11,8 @@ import multer from 'multer';
 import { expect } from 'chai';
 import { describe, it } from 'mocha';
 import {
-  type ASTVisitor,
-  type ValidationContext,
+  ASTVisitor,
+  ValidationContext,
```
IvanGoncharov, line 15: Please extract into TS type imports?

src/__tests__/http-test.ts
IvanGoncharov: > I think adding the method to the existing chain makes the most sense. Off-hand, I'm not sure why that wasn't done like that in the first place.

Agree, make total sense 👍

src/__tests__/http-test.ts
IvanGoncharov: Also it strange because `istanbul` has coverage for those lines.

src/__tests__/http-test.ts
danielrearden: I'm not clear on what you mean. Do you want these as a separate block of imports?

IvanGoncharov: I mean `import type {` syntax supported by TS.

src/__tests__/http-test.ts
```diff
@@ -963,8 +957,8 @@ function runTests(server: Server) {
       expect(JSON.parse(response.text)).to.deep.equal({
         data: {
           uploadFile: {
-            originalname: 'http-test.js',
-            mimetype: 'application/javascript',
+            originalname: 'http-test.ts',
+            mimetype: 'video/mp2t',
```
IvanGoncharov, line 961: Don't really like this :(
Does multer supports explicitly specifying MIME type?

src/__tests__/http-test.ts
```diff
@@ -2144,7 +2142,7 @@ function runTests(server: Server) {
         graphqlHTTP({
           schema: TestSchema,
           customFormatErrorFn: () => null,
-          extensions({ result }) {
+          extensions({ result }: { result: any }) {
```
IvanGoncharov, line 2145: Why do you need this cast?

src/__tests__/http-test.ts
danielrearden: Oh 🤦

src/__tests__/http-test.ts
danielrearden: I think it's just inferred from the extension. We could set up a file just for testing uploads (i.e. `fixtures/test.txt`).

src/__tests__/usage-test.ts
```diff
@@ -11,7 +9,7 @@ import { graphqlHTTP } from '../index';
 describe('Useful errors when incorrectly used', () => {
   it('requires an option factory function', () => {
     expect(() => {
-      // $DisableFlowOnNegativeTest
+      // @ts-expect-error should return runtime error
```
IvanGoncharov, line 12: These messages don't really bring any value so just change eslint config to not require a message.

src/__tests__/http-test.ts
IvanGoncharov: Also, can you do this as separate PR for all code in this repo
https://github.com/gajus/eslint-plugin-flowtype#eslint-plugin-flowtype-rules-type-import-style

src/__tests__/http-test.ts
```diff
@@ -96,11 +92,11 @@ describe('GraphQL-HTTP tests for connect', () => {
...
       request: () => supertest(app),
-      use: (...args) => app.use(...args),
+      use: app.use.bind(app),
       // Connect only likes using app.use.
-      get: (...args) => app.use(...args),
-      put: (...args) => app.use(...args),
-      post: (...args) => app.use(...args),
+      get: app.use.bind(app),
+      put: app.use.bind(app),
+      post: app.use.bind(app),
```
IvanGoncharov, line 99: Can you please extract all stuff that changes code beyond typings into separate PR so I can merge it in Flow.
Also, the move of `write`.

src/__tests__/http-test.ts
IvanGoncharov: This is also not a type change and can be done before TS convertion

src/__tests__/http-test.ts
danielrearden: 👍 Already extracted the `.write` changes earlier this morning: https://github.com/graphql/express-graphql/pull/641

src/__tests__/http-test.ts
```diff
@@ -1,6 +1,4 @@
-// @flow strict
-
-import zlib from 'zlib';
+import { gzipSync, deflateSync } from 'zlib';
```
IvanGoncharov, line 1: Same here, it can be done before TS conversion.

src/__tests__/http-test.ts
danielrearden: `result` is unknown so it cannot be spread

src/__tests__/http-test.ts
danielrearden: I just reverted this change

src/__tests__/http-test.ts
danielrearden: @IvanGoncharov So apparently the reason we wrote the tests this way to begin with is because [write returns a boolean instead of `this`](https://github.com/visionmedia/supertest/issues/433). So it can't be chained like the other methods. This also means that `@types/superagent` is wrong :(

I could just add the `await` like I originally did and that would resolve the linter issue. Do you have a preference for how we address this?

src/__tests__/http-test.ts
danielrearden: @IvanGoncharov Fixed in a separate PR

src/__tests__/http-test.ts
danielrearden: @IvanGoncharov ping

IvanGoncharov: @danielrearden Can you open PR against `supertest` to fix this?
I tried to look for a better library but it sims there nothing better ATM.

Since it's only a few places like that I'm totally ok with adding eslint-disable comments and reference supertest's PR.

src/__tests__/http-test.ts
IvanGoncharov: @danielrearden Please make a separate PR that attach with Blob as second argument:
https://developer.mozilla.org/en-US/docs/Web/API/Blob

src/__tests__/http-test.ts
IvanGoncharov: @danielrearden Can you please update result to be an `ExecutionResult` in a separate PR?

src/__tests__/http-test.ts
```diff
@@ -28,14 +26,14 @@ import {
...
-  request: () => any,
-  use: (...args: Array<mixed>) => any,
-|};
+type Middleware = (req: any, res: any, next: () => void) => any;
+type Server = () => {
+  request: () => supertest.SuperTest<supertest.Test>;
+  use: (middleware: Middleware) => any;
+  get: (path: string, middleware: Middleware) => any;
+  post: (path: string, middleware: Middleware) => any;
+  put: (path: string, middleware: Middleware) => any;
```
IvanGoncharov, line 35: @danielrearden Do we use return values anywhere?
If not it can be changed to unknown.
I just grep from time to time for `any` to remove workarounds inside typings so would like to minimize the usage of `any` in actual code.

IvanGoncharov:
@danielrearden Ready to merge.
Can you please rebase and fix the tests?

PR #641, opened 2020-06-18
Fix supertest usage inside tests

danielrearden:


IvanGoncharov:
@danielrearden Something wrong here, since GitHub actions stuck on test.
Can you please investigate?

PR #643, opened 2020-06-20
Fix middleware thunk test

danielrearden:


src/__tests__/http-test.js
```diff
@@ -1117,12 +1117,8 @@ function runTests(server: Server) {
...
       urlString(),
       graphqlHTTP((req, res) => {
-        if (req) {
-          hasRequest = true;
-        }
-        if (res) {
-          hasResponse = true;
-        }
+        hasRequest = Boolean(req);
+        hasResponse = Boolean(res);
```
IvanGoncharov, line 1121: Don't really like such conversions.
You can do:
```js
let seenRequest;

// ...
seenRequest = req;

// ...
expect(seenRequest) // ...
// Not sure how to use chai to check if it's requested, AFAIK it should be an instance of some class.
```

IvanGoncharov:
@danielrearden You doing this to increase coverage?

danielrearden:
@IvanGoncharov This fixes the issue we discussed [here](https://github.com/graphql/express-graphql/pull/640#discussion_r442084669). ESLint will complain about an unnecessary condition because the type indicates that `req` and `res` should always be defined.

By the way, as far as I can tell, istanbul is wrong -- the "else" branch is never actually hit.

IvanGoncharov:
@danielrearden Thanks for the detailed explanation 👍

PR #646, opened 2020-06-27
Fix RequestInfo typings

danielrearden:
There were two other properties that I missed as well 😓

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -171,7 +171,7 @@ export type RequestInfo = {|
   /**
    * The result of executing the operation.
    */
-  result: ExecutionResult | null,
+  result: ExecutionResult,
 
   /**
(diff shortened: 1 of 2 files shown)


src/index.js
```diff
@@ -171,7 +171,7 @@ export type RequestInfo = {|
   /**
    * The result of executing the operation.
    */
-  result: ExecutionResult | null,
+  result: ExecutionResult,
```
danielrearden, line 174: Removed the null here because `result` could never be null (see [here](https://github.com/graphql/express-graphql/blob/master/src/index.js#L330))

IvanGoncharov: Thanks for comment 👍 
It's very helpful.

PR #648, opened 2020-07-01
Update restify

danielrearden:
This was a fun issue to track down. The tests that started failing when `restify` was updated were actually for `connect`. It turns out [Restify monkey patches the request and response objects](https://github.com/restify/node-restify/blob/e648d491151484f17263c6774678f1e7ac2fa188/lib/response.js#L33). So when the connect tests ran, they ended up using the patched response methods and failed.

The solution was to delay importing the restify module until after all the other tests ran. Most of the changes to the test file just reflect that the callback passed to `runTests` is now async because `import` is async.

### DIFF
--- a/src/__tests__/http-test.ts
+++ b/src/__tests__/http-test.ts
@@ -1,8 +1,8 @@
 import zlib from 'zlib';
 
+import type { Server as Restify } from 'restify';
 import connect from 'connect';
 import express from 'express';
-import restify from 'restify';
 import supertest from 'supertest';
 import bodyParser from 'body-parser';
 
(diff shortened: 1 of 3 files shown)


src/__tests__/http-test.ts
```diff
@@ -123,7 +123,8 @@ describe('GraphQL-HTTP tests for express', () => {
 });
 
 describe('GraphQL-HTTP tests for restify', () => {
-  runTests(() => {
+  runTests(async () => {
+    const restify = await import('restify');
```
IvanGoncharov, line 127: @danielrearden Can you move it inside `describe` and make only `describe` async?

PR #658, opened 2020-07-03
Convert remaining files to TypeScript

danielrearden:


package.json
```diff
@@ -52,10 +59,13 @@
     "@babel/plugin-transform-flow-strip-types": "7.10.4",
     "@babel/preset-env": "7.10.4",
     "@babel/register": "7.10.4",
+    "@types/accepts": "^1.3.5",
```
IvanGoncharov, line 62: Please use `-E` flag for `npm install`
```suggestion
    "@types/accepts": "1.3.5",
```

resources/load-staticly-from-npm.js
```diff
@@ -15,23 +15,20 @@ const fs = require('fs');
...
+      ) {
+        const npmPath = node.arguments[0].text;
+        const filePath = require.resolve(npmPath);
+        const content = fs.readFileSync(filePath, 'utf-8');
+        return ts.createStringLiteral(content);
+      }
+    }
+    return ts.visitEachChild(node, visit, context);
   };
+  return visit;
```
IvanGoncharov, line 33: Why you can't just return arrow function?

package.json
danielrearden: Thanks sorry keep forgetting ☹️

resources/load-staticly-from-npm.js
danielrearden: It's recursive. I'll change it to a regular function.

IvanGoncharov: Ok, I see, thanks for the explanation 👍

examples/index.js
```diff
@@ -1,7 +1,9 @@
-import express from 'express';
-import { buildSchema } from 'graphql';
+/* eslint-disable import/unambiguous, node/no-unpublished-require */
 
-import { graphqlHTTP } from '../src';
+const express = require('express');
+const { buildSchema } = require('graphql');
+
+const { graphqlHTTP } = require('../src');
```
IvanGoncharov, line 6: Please switch it `index.ts` since you running it with `ts-node` anyway.

resources/build.js
```diff
@@ -6,29 +6,37 @@ const fs = require('fs');
...
+    process.cwd(),
+  );
+  const program = ts.createProgram({
+    rootNames: srcFiles.map((filepath) => path.join('./src', filepath)),
+    options,
+  });
+  program.emit(undefined, undefined, undefined, undefined, {
+    after: [transformLoadFileStaticallyFromNPM],
+  });
+  exec('npx downlevel-dts ./dist ./dist/ts3.4');
```
IvanGoncharov, line 39: Can you use it as a library?

resources/build.js
```diff
@@ -6,29 +6,37 @@ const fs = require('fs');
...
-  }
+  const { options } = ts.convertCompilerOptionsFromJson(
+    tsConfig.compilerOptions,
+    process.cwd(),
+  );
+  const program = ts.createProgram({
+    rootNames: srcFiles.map((filepath) => path.join('./src', filepath)),
+    options,
+  });
+  program.emit(undefined, undefined, undefined, undefined, {
```
IvanGoncharov, line 36: Wow, TS has a strange API

resources/load-statically-from-npm.js
```diff
@@ -4,9 +4,9 @@
 
 const fs = require('fs');
 
+const ts = require('typescript');
+
 /**
- * Eliminates function call to `invariant` if the condition is met.
- *
```
IvanGoncharov, line 9: 😄

src/index.ts
```diff
@@ -31,9 +29,12 @@ import type { GraphiQLOptions, GraphiQLData } from './renderGraphiQL';
...
 import { renderGraphiQL } from './renderGraphiQL';
 
-type $Request = IncomingMessage;
-type $Response = ServerResponse & {| json?: (data: mixed) => void |};
+type Request = IncomingMessage;
+type Response = ServerResponse & { json?: (data: unknown) => void };
 type MaybePromise<T> = Promise<T> | T;
+type Result = Pick<ExecutionResult, 'data' | 'extensions'> & {
+  errors?: ReadonlyArray<any>;
+};
```
IvanGoncharov, line 37: Why you need a custom result?

src/index.ts
```diff
@@ -124,86 +125,87 @@ export type OptionsData = {|
...
 
-type Middleware = (request: $Request, response: $Response) => Promise<void>;
+type Middleware = (request: Request, response: Response) => Promise<void>;
 
 /**
  * Middleware for express; takes an options object or function as input to
  * configure behavior, and returns an express middleware.
  */
 export function graphqlHTTP(options: Options): Middleware {
+  // eslint-disable-next-line @typescript-eslint/no-unnecessary-condition
```
IvanGoncharov, line 188: disable it inside `eslint` config with `FIXME` comment.

src/index.ts
danielrearden: `errors` in `ExecutionResult` is typed as `ReadonlyArray<GraphQLError>` but since we allow the errors in whichever way, the actual type needs to be `ReadonlyArray<any>`

resources/build.js
danielrearden: no kidding

src/index.ts
```diff
@@ -362,8 +365,10 @@ export function graphqlHTTP(options: Options): Middleware {
       response.statusCode = error.status ?? 500;
 
       if (error.headers != null) {
-        for (const [key, value] of Object.entries(error.headers)) {
-          (response: any).setHeader(key, value);
+        for (const [key, value] of Object.entries(
+          error.headers as { [key: string]: string },
```
IvanGoncharov, line 369: Please type `error` with `HttpError | Error`: https://github.com/DefinitelyTyped/DefinitelyTyped/blob/491f394a3ee3432d00284bbc204e775e3dcc5ca5/types/http-errors/index.d.ts#L17

src/index.ts
IvanGoncharov: aaa now I see why you need it because `errors` both `GraphQLError` and JSON object.

examples/index.ts
```diff
@@ -21,7 +21,7 @@ app.use(
   graphqlHTTP({
     schema,
     rootValue,
-    graphiql: { headerEditorEnabled: true },
```
IvanGoncharov, line 24: Can you please keep it?

resources/load-statically-from-npm.js
```diff
@@ -4,9 +4,9 @@
 
 const fs = require('fs');
 
+const ts = require('typescript');
+
```
IvanGoncharov, line 8: Ideally, we shouldn't require typescript since our typescript can conflict with typescript that is calling transformation.
It looks like context contain factory for creating nodes:
https://github.com/microsoft/TypeScript/blob/eb3645f16b66ed05110ea957c77773cea739bae8/src/compiler/transformers/es2019.ts#L4
Not sure but maybe it also allows accessing predicate functions.

src/index.ts
```diff
@@ -191,21 +189,23 @@ export function graphqlHTTP(options: Options): Middleware {
...
   ): Promise<void> {
     // Higher scoped variables are referred to at various stages in the asynchronous state machine below.
-    let params: GraphQLParams;
+    let params: GraphQLParams | undefined;
     let showGraphiQL = false;
     let graphiqlOptions;
-    let formatErrorFn = formatError;
+    let formatErrorFn: (
+      error: GraphQLError,
+    ) => GraphQLFormattedError = formatError;
```
IvanGoncharov, line 201: Can it be `typeof formatError`?

danielrearden: Thanks. Actually, `let formatErrorFn = formatError` is sufficient now. Had to specify the type before only because of differences in the return types.

yaacovCR:
!!!!!

PR #660, opened 2020-07-04
Fix import inside example

danielrearden:
This example won't be runnable anymore since a JS file can't import anything directly from `src`. We could make it a TS file, or just disable the lint warnings, in which case we might as well change the import to the actual library name as well.

### DIFF
--- a/examples/index.js
+++ b/examples/index.js
@@ -1,7 +1,8 @@
 import express from 'express';
 import { buildSchema } from 'graphql';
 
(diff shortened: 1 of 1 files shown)


IvanGoncharov:
@danielrearden We actually need this functionality to do `npm start` and manually test GraphiQL is working properly. So let's just migrate it to TS and this can be done in this version ahead of the main TS conversion.

PR #677, opened 2020-07-08
Improve TS type coverage

IvanGoncharov:


src/index.ts
```diff
@@ -466,8 +465,7 @@ export async function getGraphQLParams(
   request: Request,
 ): Promise<GraphQLParams> {
   // `url` is always defined for IncomingMessage coming from http.Server
-  // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
-  const urlData = new URLSearchParams(request.url!.split('?')[1]);
+  const urlData = new URLSearchParams(request.url?.split('?')[1]);
```
danielrearden, line 468: `request.url` should always be defined in the context we're using it, so I'm not sure how the coverage isn't dropping with the use of conditional chaining here :/

src/parseBody.ts
```diff
@@ -6,12 +6,9 @@ import querystring from 'querystring';
 import getBody from 'raw-body';
 import httpError from 'http-errors';
 import contentType from 'content-type';
+import type { ParsedMediaType } from 'content-type';
```
danielrearden, line 9: dang should have added a FIXME before, thanks

src/index.ts
IvanGoncharov: @danielrearden Thanks, corrected `Request` type to reflect that.

PR #696, opened 2020-09-13
Allow custom handling of runtime query errors

### LINKED ISSUE #427: Queries returning only errors are forced to be a 500

MatthiasKunnen:
Since the earlier PRs were either not ready or not flexible enough I decided to take another shot at this issue. 

Closes #427.

Replaces the following PRs:
Closes #471. 
Closes #443.

### DIFF
--- a/README.md
+++ b/README.md
@@ -123,6 +123,10 @@ The `graphqlHTTP` function accepts the following options:
 - **`formatError`**: is deprecated and replaced by `customFormatErrorFn`. It will be
   removed in version 1.0.0.
 
+- **`handleRuntimeQueryErrorFn`**: An optional function which can be used to change the status code
+  or headers of the response in case of a runtime query error. By default, the status code is set to
+  `500`.
+
 In addition to an object defining each option, options can also be provided as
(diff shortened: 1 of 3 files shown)


anthonydito:
I really like this change. Many "batteries included" hosting solutions such as AWS Elastic Beanstalk use the status code for monitoring health. Being able to distinguish between a user error and a network error in the status code would be awesome to make this more accurate.

acao (approved):
thank you @MatthiasKunnen! implementation, tests and documentation all look good to me.

nclsndr:
That's a great addition to the options! Thx you guys! When do you plan to release it?

PR #698, opened 2020-10-01
Validate static schema on middleware mount

ValeriaVG:
**Problem:**
Static schema only gets validated on the first request, which causes server to successfully start and pass health checks 

**Solution:**
If static schema is provided - validate it on mount and throw an error if validation fails

### DIFF
--- a/src/__tests__/usage-test.ts
+++ b/src/__tests__/usage-test.ts
@@ -117,4 +117,17 @@ describe('Useful errors when incorrectly used', () => {
       ],
     });
   });
+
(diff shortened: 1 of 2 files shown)


IvanGoncharov:
@ValeriaVG Totally agree that it should be done 👍 
One issue that I see is that it would be extremely hard to debug schema validation errors.
As a general rule schema should be validated once in user code, we plan to add something like `GraphQLFinalizedSchema` (validated and frozen) in `graphql-js` and force users to call something like `schema.finalize()` in their code.

I will keep this PR as a bug report until it is implemented in the next version of `express-graphql`.

PR #699, opened 2020-10-08
Fix type for customFormatErrorFn

hokaccha:
`customFormatErrorFn` may be passed `HttpError` as well as `GraphQLError`. So the argument type of `customFormatErrorFn` should be `GraphQLError | HttpError`.

## Example

```typescript
import { graphqlHTTP } from "express-graphql";
import express from "express";
import { buildSchema, formatError, GraphQLError, GraphQLFormattedError } from "graphql";
import { HttpError } from "http-errors";

const app = express();

const schema = buildSchema(`
  type Query {
    hello: String
  }
`);

app.use(
  "/graphql",
  graphqlHTTP(() => {
    return {
      schema,
      customFormatErrorFn: (err: GraphQLError): GraphQLFormattedError => {
        (err as any).extensions = {
          name: err.name,
          isHttpError: err instanceof HttpError,
        };
        return formatError(err);
      },
    };
  })
);

app.listen(8000);
```

```
$ curl -s -X POST http://localhost:8000/graphql | jq
{
  "errors": [
    {
      "message": "Must provide query string.",
      "extensions": {
        "name": "BadRequestError",
        "isHttpError": true
      }
    }
  ]
}
```

### DIFF
--- a/package.json
+++ b/package.json
@@ -50,6 +50,7 @@
     "start": "node -r ./resources/register.js examples/index.ts"
   },
   "dependencies": {
+    "@types/http-errors": "^1.8.0",
     "accepts": "^1.3.7",
     "content-type": "^1.0.4",
     "http-errors": "1.8.0",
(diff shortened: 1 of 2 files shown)


IvanGoncharov:
@hokaccha Thanks for the detailed report 👍 
I change code to always wrap HTTP errors in `GraphQLError` in #716

hokaccha:
Thanks!

PR #700, opened 2020-10-12
👷🏽‍♀️ add dependabot

A-Tokyo:
Adds dependabot with monthly checks to keep the packages up to date.


IvanGoncharov:
@A-Tokyo Thanks for the PR 👍 
But we prefer to manually update our dependencies.

PR #703, opened 2020-10-16
fix(variables) allow to send empty string in variables

ctbui:
Since this commit https://github.com/graphql/express-graphql/commit/ac9e1a9ca65e66f5c970028e7eae75a960192434
we use URLSearchParams which transform "?query=any&variables=" to variables="" 

````
const urlSearchParams = new URLSearchParams(' /graphql?query=any&variables=')
urlSearchParams.get('variables') // return ""
````

And now we have "Variables are invalid JSON." for empty variables
``````
if (typeof variables === 'string') {
    try {
      variables = JSON.parse(variables);
    } catch (error) {
      throw httpError(400, 'Variables are invalid JSON.');
    }
  } else if (typeof variables !== 'object') {
    variables = null;
  }
``````

Maybe we should allow empty string

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -479,7 +479,7 @@ export async function getGraphQLParams(
   let variables = (urlData.get('variables') ?? bodyData.variables) as {
     readonly [name: string]: unknown;
   } | null;
-  if (typeof variables === 'string') {
+  if (variables && typeof variables === 'string') {
     try {
       variables = JSON.parse(variables);
     } catch (error) {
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: src/index.ts [hunk 1]
```diff
@@ -479,7 +479,7 @@ export async function getGraphQLParams(
   let variables = (urlData.get('variables') ?? bodyData.variables) as {
     readonly [name: string]: unknown;
   } | null;
-  if (typeof variables === 'string') {
+  if (variables && typeof variables === 'string') {
```
IvanGoncharov, line 482:
```suggestion
  if (typeof variables === 'string' && variables !== '') {
```

IvanGoncharov:
@ctbui Not sure we should support this edge case unless we broke something major.
Can you provide an example of what was broken by this change?

Issue #675, opened 2020-07-07
Middleware refactor


danielrearden:
I'm opening this issue so we can further discuss approaches to A) splitting the middleware into separate functions and B) supporting additional frameworks. Here are some initial thoughts:

**Goals**
* Provide reference implementation of [GraphQL over HTTP spec](https://github.com/APIs-guru/graphql-over-http)
* Support a wider range of HTTP frameworks
* Empower consumers of the library to more easily hook into the different parts of the “request pipeline”

**Questions**
* Which HTTP libraries, frameworks or services do we want to explicitly support?
  * https://www.npmjs.com/package/express
  * https://www.npmjs.com/package/connect
  * https://www.npmjs.com/package/restify
  * https://www.npmjs.com/package/@hapi/hapi
  * https://www.npmjs.com/package/koa
  * https://www.npmjs.com/package/fastify
  * https://www.npmjs.com/package/micro
  * https://www.npmjs.com/package/sails
  * https://www.npmjs.com/package/adonis-framework
  * https://www.npmjs.com/package/derby
  * https://www.npmjs.com/package/socket.io
  * https://www.npmjs.com/package/ws
  * https://www.npmjs.com/package/aws-lambda
  * Azure Functions
  * Google Cloud Functions
  * https://deno.land/x/oak
  * Note: For serverless environments, it may be possible to just use a existing framework like express and wrap it so explicit support for serverless may not be necessary
* What features do we want to eventually support? Any potential changes to the API should be considered in the context of these features’ eventual implementation.
  * Request batching (https://github.com/graphql/express-graphql/issues/428)
  * Subscriptions (https://github.com/graphql/express-graphql/issues/390)
  * @defer and @stream directives (https://github.com/graphql/express-graphql/pull/583)
* Some features should not necessarily be a part of this library, but we want to empower the community to easily create middleware to implement them.
  * Server-side fragments (https://github.com/graphql/express-graphql/issues/575)
  * Caching (https://github.com/graphql/express-graphql/issues/561)
  * Returning the response in some format other than JSON (https://github.com/graphql/express-graphql/issues/556)
  * Custom error handling
  * Persisted queries
  * Tracing
  * Logging
  * File upload
* What steps in the “request pipeline” should be split into individual middleware?
  * Parse
  * Validate
  * Execute
  * Format response
* How do we persist state between middleware? For example, does each supported framework allow us to arbitrarily mutate the incoming request object?
* How do we account for differences between frameworks?
[...]

PR #730, opened 2020-12-12
Allow custom graphql params extraction from request

TrySound:
In my app I'm trying to move away from express. I implemented own
body parser to not pollute `request` object. Though in this case
express-graphql is trying to parse body by itself which leads to hanging
server because there are no more events on request.

In this diff I added a way to custom extracting params from request.
It would look better in options though options getter depends on parsed
options.

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -195,7 +195,16 @@ type Middleware = (request: Request, response: Response) => Promise<void>;
  * Middleware for express; takes an options object or function as input to
  * configure behavior, and returns an express middleware.
  */
-export function graphqlHTTP(options: Options): Middleware {
+export function graphqlHTTP(
+  options: Options,
+  /**
+   * An optional function which will get params from request instead of
+   * default getGraphQLParams
+   */
+  customGraphQLParamsFn?: (
+    request: Request,
+  ) => GraphQLParams | Promise<GraphQLParams>,
+): Middleware {
   devAssert(options != null, 'GraphQL middleware requires options.');
 
   return async function graphqlMiddleware(
@@ -210,11 +219,12 @@ export function graphqlHTTP(options: Options): Middleware {
     let pretty = false;
     let result: ExecutionResult;
     let optionsData: OptionsData | undefined;
+    const graphQLParamsFn = customGraphQLParamsFn ?? getGraphQLParams;
 
     try {
       // Parse the Request to get GraphQL request parameters.
       try {
-        params = await getGraphQLParams(request);
+        params = await graphQLParamsFn(request);
       } catch (error: unknown) {
         // When we failed to parse the GraphQL parameters, we still need to get
         // the options object, so make an options call to resolve just that.
--- a/package-lock.json
+++ b/package-lock.json
@@ -1003,41 +1003,6 @@
       "integrity": "sha512-+q/t7Ekv1EDY2l6Gda6LLiX14rU9TV20Wa3ofeQmwPFZbOMo9DXrLbOjFaaclkXKWidIaopwAObQDqwWtGUjqg==",
       "dev": true
     },
-    "babel-plugin-emotion": {
-      "version": "10.0.33",
-      "resolved": "https://registry.npmjs.org/babel-plugin-emotion/-/babel-plugin-emotion-10.0.33.tgz",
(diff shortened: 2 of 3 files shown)


TrySound:
cc @IvanGoncharov

acao:
excellent idea for decoupling from express.

after a quick review:
- should there be a change to package lock without dependency changes?
- potentially the function could be named more descriptively? also i think “custom” is a given in this case

acao:
also i think it makes sense to append this to the options interface rather than introducing a second argument

TrySound:
Hi @acao

> also i think it makes sense to append this to the options interface rather than introducing a second argument

> It would look better in options though options getter depends on parsed
options.

See here https://github.com/graphql/express-graphql/blob/master/src/index.ts#L231

> should there be a change to package lock without dependency changes?

Probably not. I just committed what `npm i` gives. I usually use yarn which does not introduce any changes unless packages upgraded explicitly.

> potentially the function could be named more descriptively? also i think “custom” is a given in this case

I don't know, maybe `customGetGraphQLParams`?

src/index.ts
```diff
@@ -195,7 +195,16 @@ type Middleware = (request: Request, response: Response) => Promise<void>;
...
  * configure behavior, and returns an express middleware.
  */
-export function graphqlHTTP(options: Options): Middleware {
+export function graphqlHTTP(
+  options: Options,
+  /**
+   * An optional function which will get params from request instead of
+   * default getGraphQLParams
+   */
+  customGraphQLParamsFn?: (
```
acao, line 204: @TrySound is there a reason you didn't want to add this to the `Options` type?

TrySound: I cannot because of options getter which depends on params. If we could get rid of it Im good to move into options

> It would look better in options though options getter depends on parsed
options.

TrySound:
@acao I figured how to move it into options. Now builtin getGraphQLParams is called only for options getter. When only object is passed first look for customGraphQLParamsFn field.

TrySound:
Hi @acao. What do you think?

acao:
@TrySound you did move the option to where I was hoping, awesome!

acao (changes requested):

src/index.ts
```diff
@@ -206,29 +214,34 @@ export function graphqlHTTP(options: Options): Middleware {
...
-          formatErrorFn;
-        throw error;
+      if (typeof options === 'function') {
+        try {
+          // Parse the Request to get GraphQL request parameters.
+          params = await getGraphQLParams(request);
+        } catch (error: unknown) {
+          // When we failed to parse the GraphQL parameters, we still need to get
+          // the options object, so make an options call to resolve just that.
+          optionsData = await options(request, response);
```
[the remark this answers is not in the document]
TrySound: It's not something I introduced in this PR. I'm trying to not break existing api. See readme
https://github.com/graphql/express-graphql#options
```js
app.use(
  '/graphql',
  graphqlHTTP(async (request, response, graphQLParams) => ({
    schema: MyGraphQLSchema,
    rootValue: await someFunctionToGetRootValue(request),
    graphiql: true,
  })),
);
```

IvanGoncharov:
@TrySound You can parse body yourself and replace body inside Request
https://github.com/graphql/express-graphql/blob/0fa7ace4f94fc25708a9f1e68a29775ceb623a25/src/parseBody.ts#L34-L38
 
Also, I'm worried that it will affect our effort to standardize the HTTP protocol for GraphQL: 
https://github.com/graphql/graphql-over-http
As a reference GraphQL over HTTP server, we shouldn't support non-standard names like `graphqlQuery`.

> In my app I'm trying to move away from express.

We also have a plan to do that so we can support Koa, AWS Lambdas, Deno frameworks, etc.
We currently blocked by TS migration of `graphql-js` but this is planned for the `1.0` release, see https://github.com/graphql/express-graphql/issues/675

TrySound:
> You can parse body yourself and replace body inside Request

Yes, I do request pollution now though it's not very clean.

> Also, I'm worried that it will affect our effort to standardize the HTTP protocol for GraphQL:

My change allow to move everything to userland and use any non-standard names. Any framework could be integrated here as well.

PR #733, opened 2021-01-13
Update README.md

andrehsu:
Remove `specifiedRules` from the README example for disabling introspection, since in the source code at `express-graphql/src/index.ts: 303`, `specifiedRules` is already imported and used


acao:
thank you!

PR #736, opened 2021-01-17
Revert "Allow custom handling of runtime query errors"

acao:
Reverts graphql/express-graphql#696

we have to be following the transport spec error codes:

https://github.com/graphql/graphql-over-http/blob/master/spec/GraphQLOverHTTP.md#status-codes

### DIFF
--- a/README.md
+++ b/README.md
@@ -173,10 +173,6 @@ The `graphqlHTTP` function accepts the following options:
 - **`formatError`**: is deprecated and replaced by `customFormatErrorFn`. It will be
   removed in version 1.0.0.
 
-- **`handleRuntimeQueryErrorFn`**: An optional function which can be used to change the status code
-  or headers of the response in case of a runtime query error. By default, the status code is set to
-  `500`.
-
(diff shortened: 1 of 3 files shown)


MatthiasKunnen:
How does #696 violate the linked spec?

meleyal:
The [current version of the spec](https://github.com/graphql/graphql-over-http/blob/40a50a351a887799b3d1550b665b2101b26dcda4/spec/GraphQLOverHTTP.md#status-codes) doesn't mention returning 500s, can this be reverted?

joshribakoff-sm:
I believe the authors of the library are misreading the spec:

> If the GraphQL response contains the {data} entry and it is {null}, then the server SHOULD reply with a 2xx status code and it is RECOMMENDED it replies with 200 status code.

My response contains a null data entry as described, and yet it still corrupts the status code.

Can we revert this?

mihairadulescu:
Apollo-Client cannot be used with the 500 status code !!!!!!!!!! FIX THIS ASAP!

PR #753, opened 2021-04-10
fix: improve types

kyarik:
**Changes:**
- Use `ValidationRule` type for `validationRule` option.
- Add missing type annotations.
- Remove unnecessary type annotations .


acao:
@kyarik looks great! thank you!

PR #764, opened 2021-06-01
chore: upgrade dependencies

acao:
must lock to `graphql-ws@4.1.3` for now
load static modules script is not compatible with any later version.
I get `ERR_PACKAGE_PATH_NOT_EXPORTED`?

### DIFF
--- a/package.json
+++ b/package.json
@@ -89,8 +89,8 @@
     "express": "4.17.1",
     "graphiql": "^1.4.1",
     "graphiql-subscriptions-fetcher": "0.0.2",
-    "graphql": "15.4.0",
-    "graphql-ws": "4.1.2",
+    "graphql": "15.5.0",
+    "graphql-ws": "4.1.3",
     "mocha": "8.2.1",
     "multer": "1.4.2",
     "nyc": "15.1.0",
@@ -102,12 +102,16 @@
     "sinon": "9.2.1",
     "subscriptions-transport-ws": "0.9.18",
     "supertest": "6.0.1",
-    "ts-node": "9.0.0",
-    "typescript": "4.1.2",
+    "ts-node": "10.0.0",
+    "typescript": "4.3.2",
     "unfetch": "4.2.0",
-    "ws": "5.2.2"
+    "ws": "7.4.6"
   },
   "peerDependencies": {
-    "graphql": "^14.7.0 || ^15.3.0"
+    "graphql": "^14.7.0 || ^15.5.0"
+  },
+  "optionalDependencies": {
+    "graphql-ws": "4.1.3",
+    "graphiql-subscriptions-fetcher": "0.0.2"
   }
 }
--- a/package-lock.json
+++ b/package-lock.json
@@ -328,6 +328,14 @@
         "graphql-ws": "^4.1.0",
         "meros": "^1.1.2",
         "subscriptions-transport-ws": "^0.9.18"
+      },
+      "dependencies": {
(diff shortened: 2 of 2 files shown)


acao:
@kyarik if you want to give this a review I'd love to have the help!

package.json
```diff
@@ -102,12 +102,16 @@
     "sinon": "9.2.1",
     "subscriptions-transport-ws": "0.9.18",
     "supertest": "6.0.1",
-    "ts-node": "9.0.0",
-    "typescript": "4.1.2",
+    "ts-node": "10.0.0",
+    "typescript": "4.3.2",
```
kyarik, line 106: Since we upgraded TypeScript, I think we should also upgrade `@typescript-eslint/eslint-plugin` and `@typescript-eslint/parser` to the latest version (`4.26.0`) to avoid the warning:
```
WARNING: You are currently running a version of TypeScript which is not officially supported by @typescript-eslint/typescript-estree.
```

At the same time, it would make sense to upgrade `eslint` to the latest `7.27.0`. :+1:

package.json
```diff
@@ -89,8 +89,8 @@
     "express": "4.17.1",
     "graphiql": "^1.4.1",
     "graphiql-subscriptions-fetcher": "0.0.2",
-    "graphql": "15.4.0",
-    "graphql-ws": "4.1.2",
+    "graphql": "15.5.0",
+    "graphql-ws": "4.1.3",
```
kyarik, line 93: Regarding locking to `graphql-ws@4.1.3`, we could bypass the `ERR_PACKAGE_PATH_NOT_EXPORTED` by not using `require.resolve`, but instead do a plain path concat:

```diff
- const filePath = require.resolve(npmPath);
+ const filePath = path.join(process.cwd(), 'node_modules', npmPath);
```

WDYT?

package.json
```diff
@@ -102,12 +102,16 @@
...
-    "ws": "5.2.2"
+    "ws": "7.4.6"
   },
   "peerDependencies": {
-    "graphql": "^14.7.0 || ^15.3.0"
+    "graphql": "^14.7.0 || ^15.5.0"
+  },
+  "optionalDependencies": {
+    "graphql-ws": "4.1.3",
+    "graphiql-subscriptions-fetcher": "0.0.2"
```
kyarik, line 115: Why are we listing these as optional dependencies? :thinking: 
It is unlikely that `npm` will fail to install them, so they will almost always be installed for the users of `express-graphql` (unless we expect users to run `npm install --no-optional`).
However, for those not using GraphQL subscriptions, these are two unnecessary dependencies. On the other hand, for those using subscriptions, it should be up to them to install these dependencies.

acao: Good point, they should be specified in the readme instesd

acao:
We still need to remove `graphiql-subscriptions-fetcher` and a few other unused modules

acao:
@kyarik addressed your feedback and I also switched to using `v1` by default, since `v0` is a vulnerable path technically. this puts us under 100% coverage because we don't even want `graphql-subscriptions-ws` in the lockfile for the same reason. we can eventually create a separate package.json for legacy integration tests, but for now i've lowered the threshold. codecov doesn't like this but oh well

acao:
ugh. these integration tests are a nice idea but they seem to be causing a lot of issues... not sure when this was added

https://github.com/graphql/express-graphql/pull/764/checks?check_run_id=3027687126

Issue #767, opened 2021-06-14
Interoperability with graphql go


flovilmart:
[GraphQL Go](https://github.com/graphql-go/graphql/blob/1a9db8859ef57c2821bbd47b0db9a1a09e617f41/graphql.go#L29) will always default the operation name to an empty string which causes this server to fail serving requests.

The bottom value is expected to be `null` and not `''` in this project.

Would it be acceptable that we change the implementation in this project to allow `operationName: ''` being interpreted as running the default operation?

The alternatives are:
- when using the go, always specify the operationName (cumbersome)
- updating the `graphql-go/graphql` project to allow operationName being nullable.

Would be happy to provide a PR with relevant tests.

acao:
I think this makes a lot of sense! `operationName` with an empty string value of `''` would just require an additional check for `operationName.length > 0`, and should be treated as an anonymous operation.

flovilmart:
Perfect! Let me take some time to fix it up!

flovilmart:
@acao I opened the above PR; let me know if there’s anything else I can do to help

flovilmart:
ping @acao ?

IvanGoncharov:
Closing since it should be resolved in `graphql-go`, see https://github.com/graphql/express-graphql/pull/768#issuecomment-933028222

PR #768, opened 2021-06-16
fix: Allow empty operation names for single queries

flovilmart:
Fixes issue #767

### DIFF
--- a/src/__tests__/http-test.ts
+++ b/src/__tests__/http-test.ts
@@ -813,6 +813,35 @@ function runTests(server: Server) {
       });
     });
 
+    it('allows POST with empty operation name', async () => {
+      const app = server();
+
+      app.post(
+        urlString(),
+        graphqlHTTP(() => ({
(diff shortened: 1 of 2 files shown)


flovilmart:
@saihaj or @acao ,  sorry to ping you here, any chance we can move forward with this patch?

saihaj:
I can’t merge this in but @IvanGoncharov can cut a release for this.

IvanGoncharov:
@flovilmart Did you report this issue to `graphql-go`?
GraphQL spec doesn't have any special treatment for the empty string as `operationName`: https://spec.graphql.org/draft/#sec-Executing-Requests
So I don't think it's appropriate to merge this PR.

flovilmart:
Thanks @IvanGoncharov, after reading the spec, it seems that the go implementation is indeed not following the specification regarding empty operation names.

I’ve raised the issue here: https://github.com/graphql-go/graphql/issues/616

PR #780, opened 2021-09-15
Allow custom handling of runtime query errors

### LINKED ISSUE #427: Queries returning only errors are forced to be a 500

### LINKED ISSUE #808: revert revert "Allow custom handling of runtime query errors"
https://github.com/graphql/express-graphql/pull/736 reverted something the community aligned behind changing, without addressing the concerns of the community. The author of said revert PR seems to be misreading their own link, since the spec seems pretty clear that responses that contain a `null` `data` property are recommended to contain status code 200, and nowhere does it say these responses must contain 500 which is @acao's claim.

Perhaps as a compromise, there can be a way to opt into (seemingly non spec compliant) behavior of forcing all errors to be 500 errors, however for all intents and purposes, this behavior completely breaks clients error handling behavior as described in detail in https://github.com/graphql/express-graphql/issues/427

```diff
diff --git a/node_modules/express-graphql/index.js b/node_modules/express-graphql/index.js
index 3bc78dc..d72a19d 100644
--- a/node_modules/express-graphql/index.js
+++ b/node_modules/express-graphql/index.js
@@ -174,14 +174,6 @@ function graphqlHTTP(options) {
                 result = { data: undefined, errors: error.graphqlErrors };
             }
         }
-        // If no data was included in the result, that indicates a runtime query
-        // error, indicate as such with a generic status code.
-        // Note: Information about the error itself will still be contained in
-        // the resulting JSON payload.
-        // https://graphql.github.io/graphql-spec/#sec-Data
-        if (response.statusCode === 200 && result.data == null) {
-            response.statusCode = 500;
-        }
         // Format any encountered errors.
         const formattedResult = {
             ...result,
```

<em>This issue body was [partially generated by patch-package](https://github.com/ds300/patch-package/issues/296).</em>

MatthiasKunnen:
Closes #427. 
Closes #808. 

**Why**  
When runtime errors occur, tracing and logging is paramount to discern between incorrect user input or a faulty backend/frontend. To this end, tracing info can be added to response headers in order for proxies/frontends to log this error and update metrics.

Without this PR, achieving this goal is only possible using dirty solutions such as [hijackresponse](https://www.npmjs.com/package/hijackresponse) since `express-graphql` finalizes the response which prevents later middleware from enhancing it.

### DIFF
--- a/README.md
+++ b/README.md
@@ -175,6 +175,9 @@ The `graphqlHTTP` function accepts the following options:
 - **`formatError`**: is deprecated and replaced by `customFormatErrorFn`. It will be
   removed in version 1.0.0.
 
+- **`handleRuntimeQueryErrorFn`**: An optional function which can be used to change the headers or status code of the response in case of a runtime query error. By default, the status code is set to `500`.
+  To ensure compatibility, make sure to take note of [the GraphQL spec regarding status codes](https://github.com/graphql/graphql-over-http/blob/master/spec/GraphQLOverHTTP.md#status-codes).
+
 In addition to an object defining each option, options can also be provided as
 a function (or async function) which returns this options object. This function
 is provided the arguments `(request, response, graphQLParams)` and is called
--- a/src/__tests__/usage-test.ts
+++ b/src/__tests__/usage-test.ts
@@ -2,7 +2,12 @@ import express from 'express';
 import request from 'supertest';
 import { expect } from 'chai';
 import { describe, it } from 'mocha';
-import { GraphQLSchema } from 'graphql';
+import {
+  GraphQLNonNull,
+  GraphQLObjectType,
+  GraphQLSchema,
+  GraphQLString,
+} from 'graphql';
 
 import { graphqlHTTP } from '../index';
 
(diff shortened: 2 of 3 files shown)


justinmchase:
I am strongly in favor of this, I definitely need to be able to customize the error code.

MatthiasKunnen:
Also closes #508

mihairadulescu:
What can we do to fix this? and have it in production ? Is there anything we can do to allow GraphQL specs 200 status code for error returning? This is fucking all clients , especially apollo client you cannot read body from the response !

saihaj:
This package is deprecated. You can read more here https://github.com/graphql/express-graphql/pull/817

PR #788, opened 2021-11-09
feat: support shouldPersistHeaders

ericlewis:
This exposes the `shouldPersistHeaders` option in GraphiQL, useful when you require authentication for introspection.


ericlewis:
Not sure how active it is over here, so... pinging @acao 💯

